PR #68, opened 2022-08-22
Fix anchor in spec links

### LINKED ISSUE #66: Extracted section target isn't quite right
Try running

```
duvet extract https://www.rfc-editor.org/rfc/rfc9114.html
```

From the generated toml file:

image

But the link won't navigate to the desired section. The expected target would be

```
target = "https://www.rfc-editor.org/rfc/rfc9114.html#section-3.1.1"
```

It seems like changing the `write_toml` function would be a quick fix for this particular case, but not sure if it's the right move to accommodate all use cases.

### LINKED ISSUE #67: Target of copied annotation isn't quite right
The target of an annotation I copied from the generated report was somewhat broken. The target I got was

```
//= https://www.rfc-editor.org/rfc/rfc9114#3.3
```

But the desired one would be

```
//= https://www.rfc-editor.org/rfc/rfc9114#section-3.3
```

Similar to #66 but probably a different problem. I was following the instructions from `duvet help report` but I couldn't find any options related to this. Am I missing anything?

eagr:
*Issue #, if available:*
generated spec links are broken
fix #66 
fix #67 

*Description of changes:*
fix the anchors


src/extract.rs
```diff
@@ -301,7 +301,7 @@ fn write_rust<W: std::io::Write>(
     section: &Section,
     features: &[Feature],
 ) -> Result<(), std::io::Error> {
-    writeln!(w, "//! {}#{}", target, section.id)?;
+    writeln!(w, "//! {}#section-{}", target, section.id)?;
```
camshaft, line 304: You'll need to check if the id is an appendix or not and use the `appendix-` prefix if so.

eagr: ah thanks for the pointer!

eagr: on second thought, do appendix sections also contain spec requirements?

src/extract.rs
```diff
@@ -351,3 +354,13 @@ fn write_toml<W: std::io::Write>(
...
 }
+
+fn anchor_prefix(id: &str) -> &str {
+    if let Some(_) = SECTION_ID.captures(id) {
+        &"section-"
+    } else if let Some(_) = APPENDIX_ID.captures(id) {
+        &"appendix-"
+    } else {
+        &""
+    }
```
camshaft, line 365: nit: regex seems like overkill here. could we just do:

```suggestion
    match id.chars().next().unwrap_or('_') {
        '0'..='9' => "section-",
        'a'..='z' | 'A'..='Z' => "appendix-",
        _ => ""
    }
```

www/src/result.js
```diff
@@ -53,11 +53,15 @@ input.annotations.forEach((anno, id) => {
     anno.isOk = anno.isComplete || anno.exception === anno.spec;
   }
 
+  function anchorPrefix(id) {
+    return /^[1-9]/.test(id) ? 'section' : 'appendix';
```
camshaft, line 57: This logic doesn't match the rust version. So we need to handle the case where the ID doesn't match `a-z`,`A-Z`.

camshaft (approved):

texastony (approved):

PR #75, opened 2022-09-02
ci: set up GitHub actions

camshaft:
*Description of changes:*

We are missing CI checks to make it easier to contribute to duvet. This change adds them and updates parts of the code to make the tests pass.


texastony (approved):
I am approving these changes to unblock this PR,
but I really do not know Rust and do not quite understand all that is happening here.

PR #76, opened 2022-09-02
fix: finish pattern state machine after iterating lines

### LINKED ISSUE #73: Annotation placed at the end of a source file is ignored
regardless of annotation type

camshaft:
*Issue #, if available:*

Fixes #73

*Description of changes:*

The source pattern currently doesn't terminate the state machine after iterating over all of the lines. This causes issues for files not terminated with an empty line.

This change fixes this behavior by feeding the state machine an empty line after it's done iterating. I've also added some snapshot tests to capture this behavior moving forward.


src/pattern.rs
```diff
@@ -106,6 +76,62 @@ enum ParserState<'a> {
...
 
+impl<'a> ParserState<'a> {
+    fn on_line(
+        &mut self,
+        path: &Path,
+        annotations: &mut AnnotationSet,
+        pattern: &Pattern,
+        line: &'a str,
+        line_no: usize,
+    ) -> Result<(), Error> {
```
texastony, line 87: Huh... @seebees , Rust looks like Dafny!

harrisonkaiser (approved):

harrisonkaiser:
My review didn't unblock this commit as I am not a member of awslabs/duvet-owners.

PR #79, opened 2022-09-08
Fix searching for quotes with indentations

### LINKED ISSUE #77: Unexpected trimming to the spec
Using the quote (copied exactly as is from the spec) for annotation

```
[[TODO]]
quote = '''
A CONNECT request MUST be constructed as follows:

*  The :method pseudo-header field is set to "CONNECT"

*  The :scheme and :path pseudo-header fields are omitted

*  The :authority pseudo-header field contains the host and port to
   connect to (equivalent to the authority-form of the request-target
   of CONNECT requests; see Section 7.1 of [HTTP]).
'''
```

raised an error.

```
specs/rfc9114/todos/4.4.toml#0:0 - quote not found in "https://www.rfc-editor.org/rfc/rfc9114#section-4.4"
```

Then I tried gradually removing the whitespace. And this works.

```
[[TODO]]
quote = '''
A CONNECT request MUST be constructed as follows:

*  The :method pseudo-header field is set to "CONNECT"
*  The :scheme and :path pseudo-header fields are omitted
*  The :authority pseudo-header field contains the host and port to connect to (equivalent to the authority-form of the request-target of CONNECT requests; see Section 7.1 of [HTTP]).
'''
```

Is there some kind of trimming going here for the spec before search? If so, is this the intended way to use quotes like this?

eagr:
*Issue #, if available:*
Fixes #77 
*Description of changes:*
Trim quotes line by line to allow arbitrary whitespace at line start for formatting.


camshaft:
Could you extract this logic into a separate function and write a unit test or two showing that it works how we want it to? Also you missed the `Spec` impl above.

eagr:
> Also you missed the `Spec` impl above.

Yeah, I noticed that. Considering the `Spec` type quotes were all generated and the #77 case wouldn't occur, I figured it might not need the fix.

src/source.rs
```diff
@@ -198,3 +198,52 @@ impl<'a> Todo<'a> {
...
 }
+
+fn normalize_quote(s: &str) -> String {
+    s
+        .lines()
+        .fold(String::new(), |mut s, l| {
+            let l = l.trim();
+            if !l.is_empty() {
+                s.push_str(l);
+                s.push(' ');
```
camshaft, line 209: Sorry I should have commented earlier, you can get rid of the last `.trim().into()` (which does an additional allocation and copy) by pushing a space at the beginning:

```rust
if s.is_empty() {
    s.push(' ');
}
s.push_str(l);
```

eagr: Ah, nice to know.  Thanks!

eagr: Emm, that also doesn't quite get there as it introduces a space at the start that also needs to be trimmed. How about this?

```rs
s
        .lines()
        .map(|l| l.trim())
        .filter(|l| !l.is_empty())
        .collect::<Vec<&str>>()
        .join(" ")
```

camshaft: You should only push a space if `s` isn't empty. That will prevent the space at the beginning.

I'm also fine with the `join` route since it's nice and readable but it _does_ introduce an allocation for the `collect::<Vec>`. Not a huge deal though.

eagr: Oh, I misread your comment. That's really clever.

camshaft (dismissed):

camshaft (approved):

PR #82, opened 2022-09-10
fix sorting of Requirements table

### LINKED ISSUE #81: Status sorting in Requirements table is broken
image

eagr:
*Issue #, if available:*
Fixes #81

*Description of changes:*
Fix table sorting by fixing the `sortComparator` functions


www/src/spec.js
```diff
@@ -80,8 +80,10 @@ export function Requirements({ requirements, showSection }) {
       valueGetter(params) {
         return params.row.section.id;
       },
-      sortComparator(v1, v2, row1, row2) {
-        return row1.row.cmp(row2.row);
```
camshaft, line 84: We should still use the `cmp` function, just update the logic there. It also might be a good idea to precompute the string we use to sort on since this is executed for basically every pair of requirements.

https://github.com/awslabs/duvet/blob/5e5386879c2d0145c5e0564df59ead54147e8e86/www/src/result.js#L75

eagr: Thanks for the heads-up. Then I realized the source of the problem was actually the `valueGetter`.

www/src/spec.js
```diff
@@ -78,10 +78,10 @@ export function Requirements({ requirements, showSection }) {
       field: "section",
       headerName: "Section",
       valueGetter(params) {
-        return params.row.section.id;
+        return params.row;
       },
-      sortComparator(v1, v2, row1, row2) {
-        return row1.row.cmp(row2.row);
+      sortComparator(v1, v2) {
+        return v2.cmp(v1);
```
camshaft, line 84: Should these be reversed?
```suggestion
        return v1.cmp(v2);
```

eagr: Yeah, I was thinking most people would want desc. ordering on the first click since the list was already in asc. order from the get-go. Just realized there's a better way to do it.

camshaft (approved):

PR #84, opened 2022-09-22
feat: add basic markdown support

### LINKED ISSUE #9: Markdown support in extract
Producing your own specification is overly complicated.
Since the tool prefers ietf several tools need to be used
to convert markdown files into this format.

It would be much nicer to just use the markdown directly.

camshaft:
Fixes #9

*Description of changes:*

This change adds support for citing markdown specifications. I've added a bunch of new tests to make sure nothing broke in the process.


seebees:
Using https://github.com/awslabs/aws-encryption-sdk-specification/blob/master/framework/aws-kms/aws-kms-key-arn.md as an example Markdown.

On `extract` the `-f MARKDOWN` MUST be passed.
However, with the change added,
it can recognize the markdown.
But, that's fine :)

However, the requirements parsing is off by "1".

The above produces the following extracted TOML

```
target = "aws-encryption-sdk-specification/framework/aws-kms/aws-kms-key-arn.md#appendix-A valid AWS KMS ARN"

# A valid AWS KMS ARN
#
# 
# A string with 5 `:` that delimit following 6 parts:
# 
# 1. MUST start with string `arn`
# 2. The partition MUST be a non-empty
# 3. The service MUST be the string `kms`
# 4. The region MUST be a non-empty string
# 5. The account MUST be a non-empty string
# 6. The resource section MUST be non-empty and MUST be split by a single `/`
#    any additional `/` are included in the resource id
#    1. The resource type MUST be either `alias` or `key`
#    2. The resource id MUST be a non-empty string
# 

[[spec]]
level = "MUST"
quote = '''
MUST start with string `arn`
2.
'''

[[spec]]
level = "MUST"
quote = '''
The partition MUST be a non-empty
3.
'''

[[spec]]
level = "MUST"
quote = '''
The service MUST be the string `kms`
4.
'''

[[spec]]
level = "MUST"
quote = '''
The region MUST be a non-empty string
5.
'''

[[spec]]
level = "MUST"
quote = '''
The account MUST be a non-empty string
6.
'''

[[spec]]
level = "MUST"
quote = '''
The resource section MUST be non-empty and MUST be split by a single `/`
any additional `/` are included in the resource id
1.
'''

[[spec]]
level = "MUST"
quote = '''
The resource section MUST be non-empty and MUST be split by a single `/`
any additional `/` are included in the resource id
1.
'''

[[spec]]
level = "MUST"
quote = '''
The resource type MUST be either `alias` or `key`
2.
'''

[[spec]]
level = "MUST"
quote = '''
The resource id MUST be a non-empty string
'''
```

camshaft:
The parser is missing list handling so it's assuming the number is part of the sentence on the previous line.

seebees:
There is some kind of corner case with `Appendix` in a title

```
/framework/caching-cmm/appendix-a-cache-entry-identifier-formulas.toml#0:0 - section "a-cache-entry-identifier-formulas" not found in "aws-encryption-sdk-specification/framework/caching-cmm.md"
source errors were found. no reports were generated
```

https://github.com/awslabs/aws-encryption-sdk-specification/blob/master/framework/caching-cmm.md#appendix-a-cache-entry-identifier-formulas

seebees:
In the report, the whole line is not highlighted.

seebees (approved):
These look good to me.
The ESDK specification parses now with Markdown.

PR #86, opened 2022-10-07
use BTreeMap for target data

### LINKED ISSUE #85: Report generation is not deterministic
The motivation is to avoid unnecessary report uploads/publishing. 

Right now, h3 *publishes* a compliance report on every push to the main branch, by pushing to the gh-pages branch. Most of the time, the reports are identical.

Can we somehow get the data that is fed into the reporter, in the way that it is deterministically serialized? Then we would be able to do something like hashing the data and comparing the hashes to determine whether to proceed with uploading/publishing.

eagr:
*Issue #, if available:*
Fixes #85 
*Description of changes:*
Use `BTreeMap` instead of `HashMap` to make report generation deterministic.

### DIFF
--- a/src/report/mod.rs
+++ b/src/report/mod.rs
@@ -12,7 +12,7 @@ use anyhow::anyhow;
 use core::fmt;
 use rayon::prelude::*;
 use std::{
-    collections::{BTreeSet, HashMap},
+    collections::{BTreeMap, BTreeSet, HashMap},
     path::PathBuf,
 };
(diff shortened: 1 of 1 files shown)


eagr:
I could confirm that the change does fix the problem with the local build. Do we need regression tests for this?

Also, when would you publish a new patch? Really look forward to it :))

camshaft (approved):

camshaft:
I'll try and get a release out soon

PR #87, opened 2022-10-07
chore: release 0.1.1

camshaft:
The change releases the recent bugfixes. I've also created a a CHANGELOG.md with [conventional-changelog](https://github.com/conventional-changelog/conventional-changelog).


texastony:
Are we going to add a CHANGELOG or something equivalent?

texastony (approved):

eagr:
The #81 issue has come back.

With fd82d47, I tried nuking node_modules, then ran `make build` and `cargo install`. The local build doesn't seem to have the problem. Maybe we need a fresh build? @camshaft

camshaft:
I did a fresh build as part of the release. Maybe it was the upgrade to `@mui/x-data-grid`?

eagr:
Just did these.

* `git clone` to get a new copy of duvet code
* `cargo +stable install --force --path .`
The resulted build had the same problem as the published version.

Then I tried rebuilding the js bundle and did cargo install again. Problem's gone in this build.

I think nuking the dependency tree and starting over might solve the problem. But I don't know exactly why. Did you see any funny messages after npm install when you did the build?

camshaft:
Yeah it must have somehow reverted... I've opened a new PR to avoid this in the future: https://github.com/awslabs/duvet/pull/88

PR #88, opened 2022-10-10
build: add publish script

camshaft:
This change adds a script to make publishing the crate easier. Before we committed the `script.js`, which could be easily forgotten to rebuild once publishing (see [comment](https://github.com/awslabs/duvet/pull/87#issuecomment-1272810416)). Now we generate the `script.js` on demand and force a new build before publishing.

#### TODO

- [x] Build `script.js` in CI


www/Makefile
```diff
@@ -1,9 +1,11 @@
-build:
-	@npm install
+public/script.js: package-lock.json $(wildcard src/*.js)
 	@npm run build
 	@awk 1 build/static/js/*.js > public/script.js
 
+package-lock.json:
+	@npm install
```
seebees, line 6: Ideally there would be a package-lock.json file,
and then this command would be `@npm ci`
to install a reproducible build from tested source.

camshaft: Seems reasonable

seebees (approved):
LGTM for now.
Really if you have a package-lock you should use `npm ci`
which will use that package-lock.json to have reproducible builds.

PR #89, opened 2022-10-10
fix: remove reduntant borrows

camshaft:
Clippy beta is now complaining about these.


texastony (approved):
I do not know Rust.
But I have faith in @camshaft and the checks for Duvet.

PR #96, opened 2022-11-18
Updates to the JS Report

seebees:
Enlarging the DataGrid to use the entire browser.
Changing the section and text column to use all available width.
Fix filtering by section
Add a requirements view that lists all requirements across all specifications.


www/src/App.js
```diff
@@ -136,4 +141,23 @@ function SectionRoute() {
...
+    stats: specifications.reduce((total, {stats}) => {
+      Object.keys(stats).forEach((statName) => {
+        const stat = total[statName] || new StatsClass()
+        total[statName] = stat;
+        stat.onStat(stats[statName]);
+      });
+      return total;
+    }, {}),
+    requirements,
+  }
```
camshaft, line 157: Can you move this to `result.js` so that we're not recomputing stats every time you load this page? Or at least cache the value.

www/src/App.js
```diff
@@ -136,4 +141,23 @@ function SectionRoute() {
...
+  specifications
+    .map((spec) => spec.requirements)
```
camshaft, line 159: This isn't doing anything

www/src/stats_class.js
```diff
@@ -0,0 +1,46 @@
+
```
camshaft, line 1: nit: just call this file `stats.js`

www/src/App.js
```diff
@@ -7,7 +7,9 @@ import { Nav } from "./nav";
 import { Spec, Stats } from "./spec";
 import { Section } from "./section";
 import { Link } from "./link";
+import { List } from "./list";
 import specifications from "./result";
+import { Stats as StatsClass } from "./stats_class";
```
camshaft, line 12: nit: Why does the name need `Class` at the end?

seebees: Line 7 import `Stats` from `"./spec"`

www/src/App.js
seebees: woops

www/src/App.js
seebees: Sure, it is not a lot of work and everything is local,
so I was not too worried, but :)

www/src/stats_class.js
seebees: sure

camshaft:
Closing this since it's a couple of years old at this point. I will be refactoring and refreshing the JS report a bit as well.

PR #104, opened 2023-01-14
chore: fix initial build

toidiu:
https://github.com/awslabs/duvet/pull/88 removed script.js and instead rebuilds it in CI. However this broke local development so I made a fix to generate `script.js` during the initial build.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,0 +1,12 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+use std::process::Command;
+
+fn main() {
+    Command::new("sh")
+        .arg("-c")
+        .arg("make www/public/script.js")
+        .output()
+        .expect("failed to execute process");
+}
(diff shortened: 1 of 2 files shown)


camshaft:
This will force customers installing the crate to rebuild the script, which isn't actually possible since we don't distribute the JS source in the crate.

toidiu:
The current build is broken when building for the first time: https://github.com/awslabs/duvet/blob/main/src/report/html.rs#L54

I was able to run `make www/public/script.js` locally to fix this and tested that the `build.rs` scripts also works.

camshaft:
Right, but that script should only be run in dev mode. By using a `build.rs` solution, it will force any installation to execute that code.

toidiu:
So its not possible to do a cargo build/test without scripy.js which seems like a broken experience. Not sure i understand dev mode vs "non-dev mode"?

Is non-dev mode essentially when its used for a project? How should we fix the dev mode then?

camshaft:
Let's close this PR for now and open an issue. What we have here won't work.

PR #118, opened 2023-09-21
specify path to the spec files

toidiu:
https://github.com/aws/s2n-tls/issues/4219

*Description of changes:*
Duvet was originally designed so that the spec files would be stored at project root. Since, ends up cluttering the codebase, some projects such as [s2n-tls](https://github.com/aws/s2n-tls/tree/main/compliance) instead store all spec related material in a separate directory. This results in unexpected behavior where the `report` command is unable to find the spec files and attempts to re-download them each time.

This PR expose a new argument `--spec-download-path` which can be used to specify the location where to download the spec files. Specifically for s2n-tls you could add `--spec-download-path 'compliance'` to the [generate script](https://github.com/aws/s2n-tls/blob/main/compliance/generate_report.sh). 

### Callout
I only exposed the spec-download-path for the `report` sub-command and not `extract` since it possible to run extract from any directory. We can always add this param if we want later on.

### Testing
I tested this locally to confirm that this works with the directory structure in s2n-tls.

### DIFF
--- a/src/extract.rs
+++ b/src/extract.rs
@@ -55,15 +55,23 @@ pub struct Extract {
     #[structopt(short, long, default_value = ".")]
     out: PathBuf,
 
+    /// Path to store the collection of spec files
+    ///
+    /// The collection of spec files are stored in a folder called `specs`. The
+    /// `specs` folder is stored in the current directory by default. Use this
+    /// argument to override the default location.
+    #[structopt(long = "spec-path")]
+    pub spec_path: Option<String>,
+
     target: TargetPath,
 }
 
 impl Extract {
     pub fn exec(&self) -> Result<(), Error> {
-        let contents = self.target.load(None)?;
+        let contents = self.target.load(self.spec_path.as_deref())?;
         let spec = self.format.parse(&contents)?;
         let sections = extract_sections(&spec);
-        let local_path = self.target.local(None);
+        let local_path = self.target.local(self.spec_path.as_deref());
 
         if self.out.extension().is_some() {
             // assume a path with an extension is a single file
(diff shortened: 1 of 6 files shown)


src/project.rs
```diff
@@ -55,6 +55,14 @@ pub struct Project {
...
     #[structopt(long = "spec-pattern")]
     spec_patterns: Vec<String>,
+
+    /// Path to store the collection of spec files
+    ///
+    /// The spec files are stored in a folder called `specs`. This folder is
+    /// stored at the root of the project by default. Use this argument to
+    /// specify a different location.
+    #[structopt(long = "spec-download-path", default_value = ".")]
+    pub spec_download_path: String,
```
toidiu, line 65: spec_path: Option<String>

src/extract.rs
```diff
@@ -60,10 +60,10 @@ pub struct Extract {
 
 impl Extract {
     pub fn exec(&self) -> Result<(), Error> {
-        let contents = self.target.load()?;
+        let contents = self.target.load(None)?;
```
toidiu, line 63: add the `spec_path` arg to extract as well

src/target.rs
```diff
@@ -103,10 +103,13 @@ impl TargetPath {
...
     }
 
-    pub fn local(&self) -> PathBuf {
+    pub fn local(&self, spec_download_path: Option<&String>) -> PathBuf {
         match self {
             Self::Url(url) => {
                 let mut path = std::env::current_dir().unwrap();
+                if let Some(path_to_spec) = spec_download_path {
+                    path.push(path_to_spec);
+                }
```
camshaft, line 112:
```suggestion
                let mut path = if let Some(spec_path) = spec_path {
                    PathBuf::from(spec_path)
                } else {
                    std::env::current_dir().unwrap()
                };
```

PR #125, opened 2024-10-21
build: add xtask scripts

camshaft:
### Description of changes:

This change adds a `cargo xtask` command for the repo for common tasks. The CI has also been updated to use these tasks.

I've also added integration tests with a few known customers with snapshots of reports to make sure we don't have regressions in behavior.


xtask/src/args.rs
```diff
@@ -0,0 +1,47 @@
...
+            None => default,
+        }
+    }
+}
+
+impl FlagExt for Option<Option<bool>> {
+    fn is_enabled(&self, default: bool) -> bool {
+        match self {
+            Some(Some(v)) => *v,
+            Some(None) => true,
```
toidiu, line 43: Why is this `true`?

xtask/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "xtask"
+edition = "2021"
+publish = false
```
toidiu, line 4: Could you include meta data: https://github.com/awslabs/duvet/blob/main/duvet/Cargo.toml#L1-L13

xtask/src/checks.rs
```diff
@@ -0,0 +1,38 @@
...
+    pub fn run(&self, sh: &Shell) -> Result {
+        crate::build::Build {
+            release: Some(Some(false)),
+        }
+        .run(sh)?;
+
+        {
+            let toolchain = format!("+{}", self.rustfmt_toolchain);
+            cmd!(sh, "cargo {toolchain} fmt --all -- --check").run()?;
+        }
```
toidiu, line 24: Why is the scope necessary?

xtask/src/args.rs
camshaft: so you can do these things

```console
duvet // default
duvet --foo // true
duvet --foo=true // true
duvet --foo=false // false
```

xtask/Cargo.toml
camshaft: since it's an internal crate it really doesn't need it

xtask/src/checks.rs
camshaft: i just didn't want to pollute the whole function body with variables that aren't needed

.github/workflows/ci.yml
```diff
@@ -107,12 +64,11 @@ jobs:
         with:
           submodules: true
 
-      - uses: actions-rs/toolchain@v1.0.7
+      - name: Install rustfmt toolchain
```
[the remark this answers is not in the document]
camshaft: This is intentional since we only use the nightly toolchain for rustfmt

xtask/src/checks.rs
```diff
@@ -0,0 +1,38 @@
+use crate::{args::FlagExt, Result};
+use clap::Parser;
+use xshell::{cmd, Shell};
+
+#[derive(Debug, Parser)]
+pub struct Checks {
+    #[clap(long)]
+    enforce_warnings: Option<Option<bool>>,
```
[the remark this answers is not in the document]
camshaft: same answer as https://github.com/awslabs/duvet/pull/125#discussion_r1813598071

.github/workflows/ci.yml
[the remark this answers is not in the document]
toidiu: . From what I can tell `xtask check` is what calls rustfmt. The rustfmt task is above while this task is udeps and only calls `xtask build`

xtask/src/args.rs
toidiu: Might be worth adding a comment here.

.github/workflows/ci.yml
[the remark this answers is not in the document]
camshaft: ah i should have looked at the diff :D. you're right!

toidiu (approved):

PR #127, opened 2024-10-24
refactor: improve IETF parser

camshaft:
### Description

This change improves the IETF RFC parser. Before making any changes, I included a test suite that parses all of the RFCs from 3000 to ~9600 and performs a snapshot test. I then incrementally made changes to the parser to make it more robust and handle more edge cases. All of these snapshots are stored to git LFS.

In the end, we have a lot better coverage over this parser now and the results have improved quite a bit since the logic is quite a bit more advanced.


WesleyRosenblum (approved):

camshaft:
looks like i merged this in #129

PR #136, opened 2024-11-13
refactor: move reference query to separate module

camshaft:
*Description of changes:*

This change moves all of the reference linking queries to a separate module from the report module.

### DIFF
--- a/duvet/src/reference.rs
+++ b/duvet/src/reference.rs
@@ -84,7 +84,7 @@ pub async fn build_references(
     let mut references = vec![];
 
     let Some(section_id) = section_id else {
-        // TODO return an warning that referring to a spec without a section doesn't make sense
+        // TODO return a warning that referring to a spec without a section doesn't make sense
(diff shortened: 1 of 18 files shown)


duvet/src/reference.rs
```diff
@@ -0,0 +1,145 @@
...
+    target: Arc<Target>,
+    spec: Arc<Specification>,
+    section_id: Option<Arc<str>>,
+    annotations: Arc<[AnnotationWithId]>,
+) -> (Arc<[Reference]>, Arc<[Error]>) {
+    let mut errors = vec![];
+    let mut references = vec![];
+
+    let Some(section_id) = section_id else {
+        // TODO return an warning that referring to a spec without a section doesn't make sense
```
WesleyRosenblum, line 87:
```suggestion
        // TODO return a warning that referring to a spec without a section doesn't make sense
```

WesleyRosenblum (approved):

PR #138, opened 2024-11-14
ci: fix cargo-deny task

camshaft:
*Description of changes:*

This change both updates the `cargo-deny` workflow dependency and fixes the warnings associated with it.

Note the tests have started to fail for other reasons. This is addressed in #139.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,7 +32,7 @@ jobs:
       - name: Evaluate definitions
         id: definitions
         run: |
-          export MSRV=$(cat rust-toolchain | awk '{$1=$1};1')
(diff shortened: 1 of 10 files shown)


.github/workflows/ci.yml
```diff
@@ -32,7 +32,7 @@ jobs:
       - name: Evaluate definitions
         id: definitions
         run: |
-          export MSRV=$(cat rust-toolchain | awk '{$1=$1};1')
+          export MSRV=$(rustup show | awk 'NF' | awk 'END{print $2}')
```
toidiu, line 35: 🪄

toidiu (approved):

PR #149, opened 2025-01-09
fix: improve text search normalization

camshaft:
*Description of changes:*

This change fixes a bug in the `text::find` module that doesn't correctly map characters to the original positions. The new version is more accurate. It also includes a fuzz test that has some checks around the normalized string having the specified properties.

### DIFF
--- a/duvet/src/text/find.rs
+++ b/duvet/src/text/find.rs
@@ -1,25 +1,45 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-use core::ops::Range;
-use triple_accel::levenshtein::levenshtein_search as text_search;
+use core::{fmt, ops::Range};
+use triple_accel::levenshtein::levenshtein_search_simd_with_opts as text_search;
 
-/// In order to make text matching a little nicer to work with, we split on any punctuation,
-/// rather than require strict matching
-static PUNCTUATION: &[char] = &[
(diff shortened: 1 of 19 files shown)


duvet/src/text/find.rs
```diff
@@ -32,58 +52,214 @@ fn fast_find(needle: &str, haystack: &str) -> Option<Range<usize>> {
 
 /// TODO we should probaly deprecate this - it's better to enforce strict matching
```
WesleyRosenblum, line 53: still believe this?

WesleyRosenblum (approved):

duvet/src/text/find.rs
camshaft: yep. that's why i've added the ability to warn on it 😄. still need to wire up the warnings though.

WesleyRosenblum: oh I read the TODO as suggesting not allowing fuzzy finds at all, so it wouldn't be a warning, it would just not match

camshaft: Yeah that's what I want to get to. But removing the fuzzy matching right now would be a breaking change so I'd like to give some buffer for codebases to fix the matches and then we can deprecate some time in the future.

WesleyRosenblum: ah I get it, so the warning would be something like "this is a fuzzy match, which is deprecated and will no longer be supported"

camshaft: Yeah that's exactly what I had in mind

PR #150, opened 2025-01-09
ci: fix clippy issues

camshaft:
*Description of changes:*

This fixes the latest clippy issues. I've also added a `.clippy.toml` file to specify the MSRV.


xtask/Cargo.toml
```diff
@@ -12,3 +12,9 @@ serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 toml = "0.8"
 xshell = "0.2"
+
+[lints.rust.unexpected_cfgs]
```
jmayclin, line 16: I'm admittedly coming in with little context, but could we add a comment about what this does and why we are doing it?

camshaft: can do!

jmayclin (approved):

PR #152, opened 2025-01-10
feat: add config file support

camshaft:
*Description of changes:*

This change adds support for config files for configuring `duvet`, rather than CLI arguments. This should standardize how customers set up their repositories/reports and avoid having one-off scripts everywhere:

* https://github.com/aws/s2n-quic/blob/main/scripts/compliance
* https://github.com/aws/s2n-tls/blob/main/compliance/generate_report.sh
* https://github.com/hyperium/h3/blob/master/ci/compliance/report.sh
* https://github.com/aws/aws-encryption-sdk-dafny/blob/c3ba5e314e1118efb1fd663c85e3d56578622b91/Makefile#L12-L31

The config location is `.duvet/config.toml`. All of the artifacts and `toml` files are stored alongside in `.duvet/`. I've included [a config file](https://github.com/awslabs/duvet/blob/dd828403ed1b6973654ba93b371ba739a21005f1/.duvet/config.toml) for the duvet repo itself:

```toml
'$schema' = "https://awslabs.github.io/duvet/config/v0.4.0.json"

[[source]]
pattern = "duvet/**/*.rs"

[report.html]
enabled = true
issue-link = "https://github.com/awslabs/duvet/issues"
blob-link = "https://github.com/awslabs/duvet/blob/${{ GITHUB_SHA || 'main' }}"

[report.json]
enabled = true
```

With a config file, generating a report no longer requires any arguments:

```console
$ duvet report
```

### DIFF
--- a/.duvet/config.toml
+++ b/.duvet/config.toml
@@ -1,0 +1,12 @@
+'$schema' = "https://awslabs.github.io/duvet/config/v0.4.0.json"
+
+[[source]]
+pattern = "duvet/**/*.rs"
+
+[report.html]
+enabled = true
+issue-link = "https://github.com/awslabs/duvet/issues"
+blob-link = "https://github.com/awslabs/duvet/blob/${{ GITHUB_SHA || 'main' }}"
+
+[report.json]
+enabled = true
--- a/config/v0.4.0.json
+++ b/config/v0.4.0.json
@@ -1,0 +1,197 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "https://awslabs.github.io/duvet/config/v0.4.0.json",
+  "title": "Duvet Configuration",
+  "type": "object",
+  "properties": {
+    "$schema": {
+      "type": [
+        "string",
+        "null"
+      ]
+    },
+    "report": {
+      "$ref": "#/definitions/Report"
+    },
+    "requirement": {
+      "type": "array",
+      "items": {
+        "$ref": "#/definitions/Requirement"
+      }
+    },
+    "source": {
+      "type": "array",
+      "items": {
+        "$ref": "#/definitions/Source"
+      }
+    },
+    "specification": {
+      "type": "array",
+      "items": {
+        "$ref": "#/definitions/Specification"
+      }
+    }
+  },
+  "additionalProperties": false,
+  "definitions": {
+    "CommentStyle": {
+      "type": "object",
+      "properties": {
+        "content": {
+          "default": "//#",
+          "type": "string"
+        },
+        "meta": {
+          "default": "//=",
+          "type": "string"
+        }
+      },
+      "additionalProperties": false
+    },
+    "DefaultType": {
+      "type": "string",
+      "enum": [
+        "implementation",
+        "spec",
+        "test",
+        "exception",
+        "todo",
+        "implication"
+      ]
+    },
+    "HtmlReport": {
+      "type": "object",
+      "properties": {
+        "blob-link": {
+          "anyOf": [
+            {
+              "$ref": "#/definitions/TemplatedString"
+            },
+            {
+              "type": "null"
+            }
+          ]
+        },
+        "enabled": {
+          "default": true,
+          "type": "boolean"
+        },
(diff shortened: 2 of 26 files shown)


config/v0.4.0.json
```diff
@@ -0,0 +1,197 @@
...
+      "enum": [
+        "ietf",
+        "markdown"
+      ]
+    },
+    "TemplatedString": {
+      "type": "string"
+    }
+  }
+}
```
WesleyRosenblum, line 197: should these have a newline at the end?

duvet-core/src/artifact.rs
```diff
@@ -0,0 +1,23 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+use std::path::Path;
+
+/// Synchronizes a value to the file system
+///
+/// When the `CI` environment variable is set, this method ensures the value matches
```
WesleyRosenblum, line 8:
```suggestion
/// When the `CI` environment variable is set, this method asserts the value matches
```

WesleyRosenblum: nit: `ensures` sounds more like it changes what is on disk to match, rather than just fail if it doesn't match

duvet/src/config/schema/v0_4_0.rs
```diff
@@ -0,0 +1,356 @@
...
+    pub fn load_requirements(
+        &self,
+        requirements: &mut Vec<config::Requirement>,
+        root: &Path,
+    ) -> Result {
+        // include several default paths
+        for pattern in [
+            ".duvet/requirements/**/*.toml",
+            ".duvet/todos/**/*.toml",
+            ".duvet/exceptions/**/*.toml",
```
WesleyRosenblum, line 59: any docs this should be mentioned in so the user knows they don't need to add these?

duvet/src/extract.rs
```diff
@@ -139,17 +193,17 @@ fn extract_section(section: &Section) -> (&Section, Vec<Feature>) {
 
         if KEY_WORDS_SET.is_match(line) {
             for (key_word, level) in KEY_WORDS.iter() {
-                for occurance in key_word.find_iter(line) {
+                for occurrence in key_word.find_iter(line) {
```
WesleyRosenblum, line 196: guess we need to add the typo check to CI

.duvet/config.toml
```diff
@@ -0,0 +1,12 @@
+'$schema' = "https://awslabs.github.io/duvet/config/v0.4.0.json"
```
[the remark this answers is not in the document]
camshaft: There's two reasons for it:

* I want config files to be versioned so we can make changes to the format over time and preserve support for older ones. I could have gone with `version` instead but, then that leads to the second point.
* This syntax is taken from JSON schema, where you can have your editor validate that your schema conforms to what we define. By using `$schema`, editors automatically download that URL and perform validation (see https://taplo.tamasfe.dev/configuration/using-schemas.html). It's not the nicest thing to look at... But it makes it so we don't have to register the schema anywhere and can just self host.

duvet-core/src/artifact.rs
camshaft: Yeah that makes sense

duvet/src/config/schema/v0_4_0.rs
camshaft: Yeah I will add that in the guide

duvet/src/extract.rs
camshaft: Haha yep. I added it to my entire and found a few issues 😁

.duvet/config.toml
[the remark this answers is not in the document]
WesleyRosenblum: Thanks for the explanation!

config/v0.4.0.json
WesleyRosenblum: I dont feel very strongly, its more just the way GitHub highlights it that bothers me :-)

WesleyRosenblum (approved):

PR #155, opened 2025-01-16
docs: set up guide

camshaft:
*Description of changes:*

This change sets up a guide to describe some of the concepts in duvet and how to configure the workflow. While it's not exhaustive, it's a lot better than what we have no, which is nothing.


guide/src/annotations.md
```diff
@@ -0,0 +1,113 @@
...
+```
+# Section
+
+The function MUST return a 64-bit integer.
+```
+
+In a strongly-typed language, this requirement is being both implemented and tested by the compiler.
+
+```rust
+//= my-spec.md#section
```
WesleyRosenblum, line 47: should this have `//= type=implication`?

guide/src/annotations.md
```diff
@@ -0,0 +1,113 @@
...
+
+The source code has defined an exception for a requirement and is explicitly choosing not to implement it. This could be for various reasons. For example, let's consider the following specification:
+
+```
+# Section
+
+Implementations MAY panic on invalid arguments.
+```
+
+In our example here, we've chosen _not_ to panic, but instead return an error. Annotations with the `exception` type are expected to provide a reason as to why the requirement is not being implemented.
```
WesleyRosenblum, line 64: `Annotations with the `exception` type are expected`

I forget if "reason" is required for this, maybe make it clear one way or the other if its optional or required

guide/src/annotations.md
camshaft: Yes! good catch :)

WesleyRosenblum (approved):

PR #175, opened 2025-05-21
Increase MSRV and enable Markdown heading attributes

petrp-aws:
### Description of changes
1. Update MSRV to 1.82 due to upstream libraries moving upwards as well. We now also capture our MSRV in `Cargo.toml` files of each crate.
2. Update test snapshots with new RFCs. There are also some very minor whitespace changes in `duvet report` tests.
3. Enable Markdown heading attributes with `Options::ENABLE_HEADING_ATTRIBUTES` in pulldown_cmark so that we can specify heading attributes such as anchor IDs like this: `# My Heading {#my-custom-anchor}`. I'm hoping this will help LLMs trace requirement annotations in the code back to the requirements doc.

### DIFF
--- a/duvet/src/specification/markdown/snapshots/duvet__specification__markdown__tests__heading_attributes__tokens.snap
+++ b/duvet/src/specification/markdown/snapshots/duvet__specification__markdown__tests__heading_attributes__tokens.snap
@@ -1,0 +1,64 @@
+---
+source: duvet/src/specification/markdown/tests.rs
+expression: "super :: tokens(& contents).collect :: < Vec < _ >> ()"
+---
+[
+    Content {
+        value: "",
+        line: 1,
+    },
+    Section {
+        id: Some(
+            "custom-id",
+        ),
+        title: "Heading with ID",
+        level: 1,
+        line: 2,
+    },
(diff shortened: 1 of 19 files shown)


duvet/src/specification/markdown/snapshots/duvet__specification__markdown__tests__heading_attributes__tokens.snap
```diff
@@ -0,0 +1,64 @@
...
+[
+    Content {
+        value: "",
+        line: 1,
+    },
+    Section {
+        id: Some(
+            "custom-id",
+        ),
+        title: "Heading with ID {#custom-id}",
```
camshaft, line 14: Is there a way to just get the title without the attributes? Otherwise this will probably cause some issues :/

petrp-aws: Yes, the newest revision doesn't put attributes in headings. The key to fixing this was to make sure that when we are parsing a heading and we get a non-text event, we move the end of the buffer to the end of the event range.

PR #183, opened 2025-11-03
chore: fix new clippy lints

jmayclin:
There are new clippy lints, which are causing the CI to fail. This PR fixes them.

I was also thoroughly stumped on the tar failure, so added a note about git lfs.


jmayclin:
I would vote to go ahead and override this merge. The mdbook failure appear to be related to a new 0.5.0-beta version that we are now pulling in, which appears to deprecate one of the fields.

camshaft (approved):

PR #207, opened 2026-02-20
feat: add json-v2 report format with roundtrip capability and stable identifiers

ajrudzitis:
### Motivation

Duvet generates compliance reports per package, but a single specification is often implemented across multiple packages. Merging reports from independent duvet report runs requires two things the existing JSON format lacks:

1. Roundtrip capability. The v1 JSON is write-only — it uses streaming macros with no deserialization support. A merge tool needs to read reports back in.
2. Stable identifiers. v1 assigns annotation IDs sequentially (0, 1, 2, …), so IDs from two independent runs collide. Merging requires IDs that are deterministic from content, not from insertion order.

### Structural shift: everything is an annotation

The v1 JSON format treats specifications, sections, annotations, and coverage status as fundamentally different concepts in separate top-level objects:
```
v1 topology:
├── specifications          # nested objects with line-by-line rendered content
│   └── sections            # contain lines with inline annotation ID references
├── annotations[]           # flat array, sequentially indexed
├── statuses{}              # per-offset coverage counts (spec, citation, test, …)
└── refs[]                  # pre-computed bitmask lookup table (512 entries)
```

The v2 format collapses this into a uniform model where specifications, sections, requirements, and implementation citations are all annotations. They differ only in type and the byte ranges they reference:
```
v2 topology:
├── repositories{}          # deduplicated repo blob_links, keyed by repo- ID
├── sources
│   ├── inline{}            # spec file contents (keyed by src- ID)
│   └── linked{}            # code files (keyed by lnk- ID, reference a repo)
└── annotations
    ├── specification{}     # full byte range of a source (keyed by spc- ID)
    ├── section{}           # sub-range within a source (keyed by sec- ID)
    ├── requirement{}       # authoring source + spec origin byte range(s), with coverage map (keyed by req- ID)
    └── impl{}              # code-side citation/test/etc., references target byte range(s) (keyed by cite- ID)
```

This means:
- **A specification** is an annotation spanning the full byte range of an inline source.
- **A section** is an annotation spanning a sub-range of that source.
- **A requirement** is an annotation declared at an authoring site (`source` — the TOML file or inline `//= type=spec` comment) that maps to one or more spec byte ranges (`origin`), with an explicit coverage map linking cite- IDs to the byte ranges they cover.
- **An impl** is a code-side annotation (citation, test, exception, etc.) that references one or more byte ranges in a target source.

Coverage is no longer a separate side-channel of per-offset counters and bitmask lookup tables. It's expressed directly on requirement annotations as `{ "cite-xxx": [{ start, end }] }`, making the relationship between implementation and specification self-describing.

Sources are also promoted to first-class entities. In v1, spec content is embedded line-by-line inside section objects and code file paths are bare strings on annotations. In v2, sources are extracted into their own map (inline for specs with full contents, linked for code files with optional repository references), and everything else references them by stable ID.

### Disjoint byte ranges

A single logical authoring site (one TOML `[[spec]]` entry or one inline `//= type=spec` comment) can match N disjoint byte ranges in the spec file whenever the matched quote spans regions the spec parser normalized away — IETF RFC page breaks, blank lines between paragraphs, repeated headers/footers. Both `RequirementAnnotation.origin` and `ImplAnnotation.target` use a shared `SourceRanges { src, ranges: Vec<ByteRange> }` shape to capture this faithfully: one logical annotation produces exactly one entry in the report, with all its matched ranges bundled together. The `req-` hash includes the full sorted range list so fragments collapse into a single ID deterministically.

Specification and section annotations continue to use `SourceRef` (a single contiguous range) because their byte extents are always continuous by construction.

#### Entity-typed deterministic IDs (duvet/src/ids.rs)

A new ids module provides FNV-1a 64-bit content-derived IDs with entity-type prefixes:

| Prefix | Entity | Hash input |
|---------|-------------------------|-----------------------------------------------|
| repo- | Repository | blob_link |
| src-  | Inline source (spec)    | contents (raw file bytes) |
| lnk-  | Linked source (code)    | file_name \0 repository_id |
| spc-  | Specification           | source_id \0 start \0 end |
| sec-  | Section                 | source_id \0 start \0 end |
| req-  | Requirement             | origin_id \0 s1 \0 e1 \0 s2 \0 e2 ... \0 source_id \0 line |
| cite- | Impl annotation         | source_id \0 line \0 target_source_id |

`req_id` accepts a slice of `(start, end)` tuples and sorts them internally so callers can't produce nondeterministic IDs from differently-ordered inputs. `spc_id` and `sec_id` share the same hash input `(source_id, start, end)` but use distinct prefixes so the seven entity types each own disjoint ID spaces. Includes property-based tests (bolero) for format invariants, determinism, separator safety, and range-list permutation invariance for `req_id`.

#### JSON v2 report schema (duvet/src/report/json_v2.rs)

A new json_v2 module implements the full v2 report format:

- ReportV2 top-level struct with version, issue_links, repositories, sources, and annotations.
- Sources split into inline (spec file contents embedded in the report) and linked (code files referencing a repository). Both `InlineSource.file_name` and `LinkedSource.file_name` store the full path as duvet saw it; consumers that want just the basename can derive it.
- Annotations split into four types: specification, section, requirement (with authoring source + spec origin ranges + coverage map), and impl (with source location + nested target ranges).
- Shared `SourceRanges { src, ranges }` type used for any pointer into a spec file that may cover disjoint regions (requirement origin, impl target). `SourceRef` retained for single-range references (specification and section annotations).
- Schema URL keys in serialized JSON (e.g., https://awslabs.github.io/duvet/v2/sources.json#inline) for future schema validation.
- Full serde Serialize/Deserialize roundtrip with deny_unknown_fields.
- `from_report_result()` converts the internal ReportResult into the v2 format. Spec references are grouped by authoring site `(origin src-, source lnk-, anno_line)` so one logical requirement yields one `RequirementAnnotation` regardless of how many disjoint fragments its quote produced. Coverage clamping iterates each origin range so a single impl reference can contribute under multiple origin ranges of the same requirement.
- Read/write helpers for JSON I/O with version validation.
- `AnnotationType` conversion uses `TryFrom` returning `Err(())` for `Spec`, making the Step 5 filter type-enforced: `let Ok(new_anno_type) = ... else { continue; }` simultaneously rejects Spec references and hoists the conversion.

#### JSON Schema (schemas/report-v2.json)

- `schemars::JsonSchema` derived (test-only) on all ReportV2 types.
- A test generates `schemas/report-v2.json` via `duvet_core::artifact::sync` — locally it writes the file, in CI it asserts the checked-in copy matches (fails the build if the schema is stale).

#### CLI integration (duvet/src/report.rs)

- New `--json-v2 <path>` CLI flag for duvet report.
- `DUVET_INTERNAL_CI_JSON_V2` environment variable for CI-driven report generation.
- TargetReport fields and the status module changed to pub(crate) visibility so the json_v2 module can access them.

#### Integration tests (xtask/src/tests.rs)

- All integration tests now generate and snapshot v2 JSON reports alongside v1.
- New *_json_v2.snap snapshot files tracked via Git LFS.


duvet/src/report/json_v2.rs
```diff
@@ -0,0 +1,1199 @@
...
+//!           "title": "<string>",
+//!           "lines": [
+//!             { "type": "plain", "content": "<string>" },
+//!             { "type": "segmented", "content": [
+//!               {
+//!                 "annotation_ids": ["<stable_id>", ...],
+//!                 "status": "<u8 bitmask>",
+//!                 "text": "<string>"
+//!               }
+//!             ]}
```
camshaft: One thing I worry about is how much this will inflate the output, especially if you have a heavily annotated source file.

I think another option could be just putting the entire string contents in a single field and then having the annotations refer to byte ranges. That would probably be the most minimal amount of overhead for source files, but I understand it can create a bit more work if you want to actually consume this too.

duvet/src/report/json_v2.rs
```diff
@@ -0,0 +1,1199 @@
...
+//!           ],
+//!           "requirements": ["<stable_id>", ...]
+//!         }
+//!       ]
+//!     }
+//!   },
+//!   "annotations": [
+//!     {
+//!       "id": "<16-char hex stable ID>",
+//!       "source": "<file path>",
```
camshaft: Something I've been wanting to do is embed source files in the output so you can more easily display them inline rather than linking externally. I think this would also help with merging if you had a top-level key of `sources` and it could include the raw contents, blob link, etc.

duvet/src/report/json_v2.rs
```diff
@@ -0,0 +1,1199 @@
...
+    if level_bits != 0 {
+        level_str = format!("level={}", AnnotationLevel::from_bits(level_bits));
+        parts.push(&level_str);
+    }
+
+    parts.join(", ")
+}
+
+/// The top-level v2 report structure.
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
```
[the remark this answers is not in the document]
ajrudzitis: Do you have an example of how that should work in the build process? I assume we want to check the schema in, and then assert during the build process that it has not changed?

duvet/src/report/json_v2.rs
```diff
@@ -0,0 +1,772 @@
...
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+pub struct AnnotationsV2 {
+    #[serde(rename = "https://awslabs.github.io/duvet/v2/annotations.json#specification")]
+    pub specification: BTreeMap<String, SpecificationAnnotation>,
+    #[serde(rename = "https://awslabs.github.io/duvet/v2/annotations.json#section")]
+    pub section: BTreeMap<String, SectionAnnotation>,
+    #[serde(rename = "https://awslabs.github.io/duvet/v2/annotations.json#requirement")]
+    pub requirement: BTreeMap<String, RequirementAnnotation>,
+    #[serde(rename = "https://awslabs.github.io/duvet/v2/annotations.json#impl")]
```
ajrudzitis: I'm not in love with the name `impl`. I think it should either be the full `implementation` or something like `citation`.

duvet/src/report/json_v2.rs
```diff
@@ -0,0 +1,772 @@
...
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+pub struct RequirementAnnotation {
+    pub source: SourceRef,
+    pub level: AnnotationLevel,
+    #[serde(default, skip_serializing_if = "BTreeMap::is_empty")]
+    pub coverage: BTreeMap<String, Vec<ByteRange>>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+pub struct ImplAnnotation {
```
ajrudzitis: Ditto on naming here.

duvet/src/report/json_v2.rs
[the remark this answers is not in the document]
ajrudzitis: Figured it out. Schema is in schemas/report-v2.json. Let me know if there is a better place to put it. We could also relocate the config schema?

duvet/src/report/json_v2.rs
WesleyRosenblum: how about `cite`?

duvet/src/report/json_v2.rs
```diff
@@ -0,0 +1,1052 @@
...
+
+            let file_name = reference.annotation.source.to_string_lossy().to_string();
+            let repo_id = resolve_repo_id(
+                reference.annotation.blob_link.as_deref(),
+                report,
+                blob_link_to_repo_id,
+            );
+
+            let lnk_key = (file_name, repo_id);
+            let lnk_id = source_to_lnk_id.get(&lnk_key).cloned().unwrap_or_default();
```
WesleyRosenblum: Could this use `expect(...)` instead of `unwrap_or_default()`? Probably better to just cause a failure if its actually required. Same for the one on line 697

xtask/src/tests.rs
```diff
@@ -296,11 +302,15 @@ impl IntegrationTest {
         let json_file = sh.read_file(&json_report)?;
         let json: serde_json::Value = serde_json::from_str(&json_file)?;
 
+        let json_v2_file = sh.read_file(&json_v2_report)?;
+        let json_v2: serde_json::Value = serde_json::from_str(&json_v2_file)?;
+
```
WesleyRosenblum, line 307: Could you add a round trip assertion here, something like:

`let _: duvet::report::json_v2::ReportV2 = serde_json::from_str(&json_v2_file)?;`

WesleyRosenblum: or have it as a unit test in the duvet source

duvet/src/report/json_v2.rs
ajrudzitis: Fixed.

xtask/src/tests.rs
ajrudzitis: Added here: https://github.com/awslabs/duvet/pull/207/changes/5f11c9b626c9f3c4c033bebfcb61cf0bb5eab838

duvet/src/report/json_v2.rs
ajrudzitis: Also done!

duvet/src/ids.rs
```diff
@@ -0,0 +1,305 @@
...
+
+/// FNV-1a 64-bit hash function.
+///
+/// Deterministic, no dependencies, sufficient for expected annotation counts.
+/// Uses the standard FNV-1a constants for 64-bit hashing.
+///
+/// Note: 64-bit hashes have a birthday-bound collision probability of ~1 in 2^32
+/// (~4 billion) at around 77k entries. This is acceptable for any realistic
+/// project, but IDs should not be assumed to be universally unique — a merge
+/// tool should detect and handle collisions.
```
seebees, line 32: This is very likely. I am just asking. `cite_annotations` has a `debug_assert_eq!`, but all of the other places silently insert the duplicates. I think I agree that this is unlikely in a single project. If it did occur, would it be better for the program to halt? Or to silently continue or what?

duvet/src/ids.rs
```diff
@@ -0,0 +1,305 @@
...
+    }
+
+    #[test]
+    fn lnk_id_known_vector() {
+        assert_eq!(lnk_id("src/lib.rs", "repo-abc123"), "lnk-944cdb82e9f1060d");
+    }
+
+    #[test]
+    fn lnk_id_no_repository() {
+        assert_eq!(lnk_id("src/lib.rs", ""), "lnk-45b7397fa826f08e");
```
seebees, line 132: It worries me that you are trying to "Make merging easy" but this creates a collision on file names. The hash argument does not fly as well here. Since `lib.rs` is a relatively likely file name in different projects.

I'm not blocking. But I think that your goal and implementation are at odds here :)

seebees (commented):
LGTM. Non-blocking comments added.
There are other things I could talk about, but I don't think that an extended back and forth is required :)

duvet/src/ids.rs
ajrudzitis: Good question. The reason cite has it is that cite is the one site where ID convergence is *expected* — multiple references intentionally collapse into one `CiteAnnotation` with merged `target.ranges`, so the assert guards against side-channel fields (`anno_type`, `level`, `comment`, …) drifting across references that should agree by construction.

The other sites are dedup, not merge:
- `repo-` and `lnk-` are structurally idempotent — the duplicate is bit-identical, so an assert would just be noise.
- `req-` is grouped by `(src_id, lnk_id, anno_line)` per target, so the `BTreeMap` insert can't see a real duplicate within a single report.
- `src-`, `spc-`, and `sec-` *do* carry side-channel fields (`file_name`, `title`/`format`, `short_name`/`long_name`) that could in principle drift, so I added `debug_assert_eq!` at those three sites in the latest push.

On halt vs. continue: I kept `debug_assert_eq!` (panics in dev/CI, no-op in release) rather than halting end-user runs. A real 64-bit FNV collision is ~1-in-2^32 and shouldn't fail a report; a logic bug producing inconsistent metadata is what we actually want to catch, and dev/CI is where we'll see it. Cross-report collision detection belongs in the merge tool, not single-report construction.

duvet/src/ids.rs
ajrudzitis: I agree, this is not as robust as it could be. Right now, the solution is for a user to provide a blob_link for their source, which will ensure the files get a strong link id. We could be more aggressive and assign an explicit id.

camshaft (dismissed):

schemas/report-v2.json
```diff
@@ -0,0 +1,372 @@
...
+          }
+        }
+      },
+      "additionalProperties": false,
+      "required": [
+        "https://awslabs.github.io/duvet/v2/annotations.json#specification",
+        "https://awslabs.github.io/duvet/v2/annotations.json#section",
+        "https://awslabs.github.io/duvet/v2/annotations.json#requirement",
+        "https://awslabs.github.io/duvet/v2/annotations.json#cite"
+      ]
```
camshaft: sorry - i just noticed this but i think we should allow for arbitrary extensions here. and i don't think any of them are strictly required either.

schemas/report-v2.json
```diff
@@ -0,0 +1,372 @@
...
+          "additionalProperties": {
+            "$ref": "#/$defs/LinkedSource"
+          }
+        }
+      },
+      "additionalProperties": false,
+      "required": [
+        "https://awslabs.github.io/duvet/v2/sources.json#inline",
+        "https://awslabs.github.io/duvet/v2/sources.json#linked"
+      ]
```
camshaft: same here - it would be nice to allow for flexibility to extend the schema with other source types.

schemas/report-v2.json
ajrudzitis: Done — pushed in 0c1a466. `AnnotationsV2` now has `additionalProperties: true` and no `required` keys; on the Rust side a `#[serde(flatten)] extensions: BTreeMap<String, serde_json::Value>` field captures unknown URL-keyed entries so they survive roundtrip through merge tooling. Added `unknown_extension_keys_roundtrip` to lock the contract in.

schemas/report-v2.json
ajrudzitis: Same treatment in 0c1a466 — `SourcesV2` is now extensible and none of `#inline`/`#linked` are required. The flattened `extensions` map preserves unknown source-type buckets verbatim, so future source types (or out-of-band tools) can ride alongside the known ones without a schema bump.

camshaft (approved):

PR #211, opened 2026-03-21
build: update reqwest to 0.13

gibbz00:
The intention is to simplify building duvet on more targets.

Upgrading request to 0.13 switches in turn the TLS provider to rustls with aws-lc as the crypto provider. This drops the requirement to have openssl and pkg-config and installed at build-time.

- [x] By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.


dougch (approved):

CarolYeh910:
Hi @gibbz00, thanks for creating this PR. The change has been merged in #225, so we can now close this one.

PR #220, opened 2026-05-14
feat: add `duvet merge` subcommand for combining v2 JSON reports

ajrudzitis:
## Summary

Adds `duvet merge`, a subcommand that combines N v2 JSON reports into a single v2 JSON report. Builds on the v2 schema's deterministic content-derived IDs (#207): same content in two inputs hashes to the same ID, so merging is a per-entity `BTreeMap` union with conflict checks on the fields that aren't part of each ID's hash.

```
duvet merge --input package-a/duvet_report_v2.json \
            --input package-b/duvet_report_v2.json \
            --output merged.json
```

## Conflict policy

Each field on every entity type falls into one of three buckets:

1. **Hash-input** (e.g. `Repository.blob_link`, `InlineSource.contents`, `RequirementAnnotation.origin`). ID equality implies field equality, so a mismatch is either a hash collision or a producer bug. The merger errors out — silently picking a side could merge structurally distinct entities. Errors are tagged "internal invariant violation".
2. **Structural conflict** (e.g. spec `title`/`format`, section `short_name`/`long_name`, requirement `level`, cite `anno_type`/`level`/`target`, extension values). Out-of-hash fields where producers must agree; mismatches hard-error.
3. **Soft drift** (cite `comment`, `feature`, `tracking_issue`, `tags`). Routinely differs across branches and isn't load-bearing for traceability — reconciled deterministically so the binary merge is commutative and associative (and the N-way fold is therefore order-independent). For the `Option<String>` fields, `Some` beats `None`, and when both sides are `Some` the lexicographically smaller value wins. `tags` becomes the sorted union of both sets. Any field that actually differed is logged as a stderr warning so the drift remains observable.

`RequirementAnnotation.coverage` is set-valued: cite-keyed range lists are unioned across inputs, then sort+dedup'd at the end so the merged output is canonical regardless of input order. `issue_links` is likewise a sorted, deduped union (the schema is a free-form `Vec<String>` with no ID, so a canonical sort order is what keeps the merge commutative).

## Tests

- 36 unit/property tests in `duvet/src/report/json_v2_merge.rs` covering every error path, soft-drift warnings, set unions, and the algebraic properties (commutativity, associativity, idempotency) — 7 of them are property tests. One test self-merges a real-world v2 snapshot.
- `merge-multi-package` integration fixture: two packages with overlapping coverage of one shared spec, merged end-to-end via the `duvet merge` CLI; output captured as insta snapshots.
- `merge-three-package` integration fixture: 3-way fold across three packages each citing one shared feature plus one distinct feature, exercising the N>2 path.

## Test plan

- [x] `cargo xtask checks` — clippy + rustfmt + typos clean
- [x] `cargo xtask test` — full suite passes (lib + integration + doc)
- [x] Self-merge of a real-world v2 snapshot is identity (`merge_real_snapshot_self_idempotent`)
- [x] Property tests assert `merge(a,b) == merge(b,a)` and `merge(merge(a,b),c) == merge(a,b,c)`

### DIFF
--- a/duvet/src/report/json_v2_merge.rs
+++ b/duvet/src/report/json_v2_merge.rs
@@ -66,6 +66,39 @@ pub fn merge_reports(reports: Vec<ReportV2>) -> crate::Result<ReportV2> {
         ));
     }
 
+    // Linked sources without a `repository` collapse to a `lnk-` ID hashed
+    // from `file_name` alone. Within a single report that's fine — paths
+    // are unique by construction. The hazard is across reports: if two
+    // unrelated projects each have a no-repo linked source for the same
+    // path (e.g. `src/lib.rs`), they share a `lnk-` ID and merging them
+    // would silently consolidate annotations from disjoint codebases.
+    //
+    // We can't tell same-project from cross-project at merge time, so we
+    // refuse the merge whenever the same no-repo `lnk-` ID appears in two
+    // or more inputs. Single-input occurrences are safe and remain allowed.
+    {
+        let mut no_repo_seen: BTreeMap<&str, usize> = BTreeMap::new();
+        for (i, report) in reports.iter().enumerate() {
+            for (lnk_id, src) in &report.sources.linked {
+                if src.repository.is_some() {
+                    continue;
+                }
+                if let Some(&prior) = no_repo_seen.get(lnk_id.as_str()) {
+                    return Err(duvet_core::error!(
+                        "linked source '{}' (file '{}') without a repository appears in inputs #{} and #{}; \
+                         merging would silently consolidate it across reports — \
+                         add a blob-link to disambiguate (see [[source]] blob-link in your duvet config)",
+                        lnk_id,
+                        src.file_name,
+                        prior,
+                        i,
+                    ));
+                }
+                no_repo_seen.insert(lnk_id, i);
+            }
+        }
+    }
+
     // Use the first input as the accumulator and fold the rest in. Because
     // every per-bucket merge operation is a BTreeMap union over content-
     // hashed IDs, the choice of seed doesn't affect the final result —
@@ -470,8 +503,8 @@ fn finalize_coverage(requirements: &mut BTreeMap<String, RequirementAnnotation>)
 mod tests {
     use super::*;
     use crate::report::json_v2::{
-        AnnotationLevel, AnnotationType, ByteRange, InlineSource, SourceLocation, SourceRanges,
-        SourceRef,
+        AnnotationLevel, AnnotationType, ByteRange, InlineSource, LinkedSource, SourceLocation,
+        SourceRanges, SourceRef,
     };
 
     fn empty() -> ReportV2 {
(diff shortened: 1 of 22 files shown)


duvet/src/report/json_v2_merge.rs
```diff
@@ -0,0 +1,1039 @@
...
+
+/// `cite-` IDs hash `(source_id, line, target_source_id)`, so `source` and
+/// `target.src` are guaranteed by ID equality. `target.ranges`,
+/// `anno_type`, and `level` are out-of-hash structural fields — mismatches
+/// here would change the meaning of the citation, so we hard-error.
+///
+/// The "soft" cite metadata (`comment`, `feature`, `tracking_issue`,
+/// `tags`) is not load-bearing for traceability and routinely drifts
+/// across branches; we keep the first value seen and warn — see
+/// [`warn_on_soft_drift`].
```
WesleyRosenblum: I think this conflicts with the comment that "the same input pair produces the same output regardless of input order."

seebees: I agree. I would make the soft drift commutative. When values disagree pick deterministically. Lexicographically smaller for strings?

duvet/src/report/json_v2_merge.rs
```diff
@@ -0,0 +1,1039 @@
...
+/// * A hash-input invariant fails — i.e. two inputs share an ID but differ
+///   on a field that goes into that ID's hash. This is either a hash
+///   collision or a producer bug; either way we refuse to silently merge.
+pub fn merge_reports(reports: Vec<ReportV2>) -> crate::Result<ReportV2> {
+    if reports.is_empty() {
+        return Err(duvet_core::error!(
+            "merge requires at least one input report"
+        ));
+    }
+
```
seebees, line 72: As discussed before if there is no linked source the annotations will conflict. I would suggest that we just error here.

```suggestion

      // Linked sources without a repository ID collapse to a lnk- ID derived
      // only from file_name, so two distinct projects' "src/lib.rs" produce
      // identical IDs and would silently consolidate. Refuse to merge such
      // reports — the user must add a blob-link to every source pattern in
      // their duvet config first.
      for (i, report) in reports.iter().enumerate() {
          for (lnk_id, src) in &report.sources.linked {
              if src.repository.is_none() {
                  return Err(duvet_core::error!(
                      "input #{} has linked source '{}' (file '{}') without a repository; \
                       merging requires every linked source to have a blob-link configured \
                       (see [[source]] blob-link in your duvet config)",
                      i, lnk_id, src.file_name,
                  ));
              }
          }
      }

```

duvet/src/report/json_v2_merge.rs
```diff
@@ -0,0 +1,1039 @@
...
+fn merge_specification(
+    out: &mut BTreeMap<String, SpecificationAnnotation>,
+    id: String,
+    anno: SpecificationAnnotation,
+) -> crate::Result {
+    match out.get(&id) {
+        Some(existing) => {
+            if existing.title != anno.title {
+                return Err(duvet_core::error!(
+                    "specification annotation '{}' has conflicting title across inputs: {:?} vs {:?}",
```
seebees: For a merge of 2 this seems pretty good. But for a merge of N does this still work?

duvet/src/report/json_v2_merge.rs
```diff
@@ -0,0 +1,1039 @@
...
+    for (id, repo) in incoming {
+        match out.get(&id) {
+            Some(existing) => {
+                // `blob_link` is the sole hash input for `repo-` IDs, so
+                // ID equality must imply blob_link equality. If it doesn't,
+                // we've hit a hash collision (or a producer bug) and have
+                // no safe way to pick between the two.
+                if existing.blob_link != repo.blob_link {
+                    return Err(duvet_core::error!(
+                        "internal invariant violation: repository '{}' has different blob_link across inputs ('{}' vs '{}'); likely hash collision or producer bug",
```
seebees: Ok... so what do I do? :) 


```suggestion
                        "internal invariant violation: repository '{}' has different blob_link across inputs ('{}' vs '{}'); likely hash collision or producer bug -- this is rare; please file a bug with both input reports attached.",
```

seebees (changes requested):

duvet/src/report/json_v2_merge.rs
ajrudzitis: Added a more deterministic ordering in c3e0104aaa5715755165dcb3573f30e8efd633c4

duvet/src/report/json_v2_merge.rs
ajrudzitis: Done in c728318a97f5ace11cd28113b6e0ef366d36a364

duvet/src/report/json_v2_merge.rs
ajrudzitis: Do you mean in terms of ux or functionality?

duvet/src/report/json_v2_merge.rs
ajrudzitis: I took a different approach in e068fcbc6d42ee9b16e7e3f3e592aece7f02c84f. We will check for collisions in linked source file names. If there is a collision and no repository id, then fail.

duvet/src/report/json_v2_merge.rs
seebees: UX. if there is an N way merge does this error message still work? trying to work it out in my mind.

duvet/src/report/json_v2_merge.rs
```diff
@@ -0,0 +1,1326 @@
...
+/// `issue_links` is an order-preserving deduped concatenation. We can't use
+/// a `BTreeMap` here because the schema is a free-form `Vec<String>` with
+/// no ID, but we still want a stable, idempotent result: appending a link
+/// already present is a no-op.
+fn merge_issue_links(out: &mut Vec<String>, incoming: Vec<String>) {
+    for link in incoming {
+        if !out.contains(&link) {
+            out.push(link);
+        }
+    }
```
seebees: This is non-commutative. Because there is no canonical iteration on `Vec`.
I think that the same soft-drift reconciliation works here too.

I think all your tests have an empty `issue_links`. I think you start with `r = empty();` everywhere and then set something for everything but issue links :).

```suggestion
      out.extend(incoming);
      out.sort();
      out.dedup();
```

I think this will show us:

```
  #[test]
  fn show_the_gap() {
      let mut a = empty();
      a.issue_links = vec!["x".into(), "y".into()];
      let mut b = empty();
      b.issue_links = vec!["y".into(), "z".into()];

      let forward = merge_reports(vec![a.clone(), b.clone()]).unwrap();
      let reverse = merge_reports(vec![b, a]).unwrap();
      assert_eq!(forward, reverse);  // fails?
  }
```

seebees (changes requested):

duvet/src/report/json_v2_merge.rs
ajrudzitis: Applied in ba7d456952cdd6baea2ccff4493d054d6df35249, along with a test.

duvet/src/report/json_v2_merge.rs
ajrudzitis: I updated the error messages in in fe56692 to indicate which file is being added to the accumulator when it fails. Because of how we're merging them, keeping track of the first place a value is seen is a lot of book keeping, which I'm not sure is worth it.

seebees (approved):
LGTM!

ajrudzitis:
Heads-up for anyone reading the squash commit message on \`main\`: two parts of the conflict-resolution description don't match what actually shipped. The PR description above has been updated; the squash message itself is frozen.

- **Soft drift** (cite \`comment\`, \`feature\`, \`tracking_issue\`, \`tags\`) is *not* first-writer-wins. It's reconciled deterministically so the binary merge is commutative and associative: for the \`Option<String>\` fields, \`Some\` beats \`None\` and the lex-min string wins when both sides are \`Some\`; \`tags\` is the sorted union. Drift is still logged to stderr.
- **\`issue_links\`** is a sorted, deduped union (not "deduplicate while preserving first-seen order") — same commutativity reason.

See \`duvet/src/report/json_v2_merge.rs\` (\`merge_soft_drift\`, \`merge_issue_links\`) for the authoritative behavior.

camshaft:
we should probably add some docs around how to use the merge command, similar to the others: https://github.com/awslabs/duvet/blob/ddfe05c452c0af40c5af6f3ae7083ee79389e4b6/xtask/src/guide.rs#L50 https://awslabs.github.io/duvet/command/init.html

PR #228, opened 2026-05-29
fix(xtask): sync RFCs via rsync instead of tarball

ajrudzitis:
The RFC-all.tar.gz endpoint now returns 404. Switch to the rsync mirror documented at rfc-editor.org/series/rfc-download/.


dougch (approved):
thanks for doing this

camshaft (approved):

PR #235, opened 2026-07-21
feat: convert JSON v2 reports to v1

ajrudzitis:
*Issue #, if available:*

N/A

*Description of changes:*

Adds an internal JSON v2 to v1 compatibility bridge for the existing web UI. The implementation remains directly invocable for integration testing:

```bash
duvet convert \
  --input report-v2.json \
  --output report-v1.json \
  [--issue-link URL] \
  [--validate-against direct-report-v1.json]
```

`convert` is intentionally hidden from `duvet --help` and is not a supported general-purpose conversion interface. Its v1 output is an intermediary for the legacy React report frontend; keeping the hidden entry point callable lets the compatibility path be exercised end to end before HTML generation consumes it internally.

The converter builds serde-backed v1 DTOs directly from `ReportV2` without reconstructing `ReportResult`. It reparses embedded IETF and Markdown specifications, verifies recorded section metadata, assigns deterministic integer annotation IDs, reconstructs line segmentation and the 256-entry refs table, recomputes requirement coverage statuses from interval unions, and resolves per-annotation blob links through v2 repositories.

Reparsing is intentionally fail-hard: conversion requires the current parser to reproduce the section metadata recorded in v2. Reports produced by parser-incompatible Duvet versions must be converted with a compatible version. This limitation remains available in `duvet convert --help` for internal users and test diagnostics. The report version is treated as an exact wire-schema revision rather than package SemVer, so the converter accepts `2.0` only until a later schema has been reviewed explicitly.

`--validate-against` compares canonical semantic models instead of raw JSON. It normalizes ordering and integer IDs, canonicalizes RFC URL variants and IETF section aliases, expands refs entries, applies effective blob-link fallback, overlays duplicate direct-v1 RFC targets, and compares specification text, sections, annotations, statuses, coverage, and metadata strictly. Requirement comments are compared through their specification segments because original authored quote formatting is not retained by v2.

*Verified canonicalization:*

Semantic line comparison coalesces adjacent spans with identical annotation sets, so arbitrary segment cuts do not create false mismatches. The canonicalizer lives in `duvet/src/report/canonical.rs` inside a `verus!` block. Its model expands spans into a per-byte sequence of annotation-set labels (`denote`), and Verus proves four things about it:

- **Implementation matches specification.** The executable `canonicalize_spans` carries `ensures output@ == canonicalize_spec(input@)`, so every theorem below is about the code that runs, not a parallel model of it.
- **Denotation preservation.** `denote(canonicalize_spec(input)) == denote(input)`: canonicalization never changes per-byte coverage.
- **Soundness.** `canonicalize_spec(l) == canonicalize_spec(r)` ⟹ `denote(l) == denote(r)`: canonicalization cannot mask a real coverage difference, so a validation *pass* can be trusted.
- **Completeness.** `denote(l) == denote(r)` ⟹ `canonicalize_spec(l) == canonicalize_spec(r)`: equal per-byte coverage always canonicalizes identically, so no difference in segmentation can produce a false *failure*. Proven through canonical-form uniqueness — canonicalization emits maximal runs (`coalesced`), and a well-formed coalesced span list is uniquely determined by its denotation (`lemma_coalesced_denotation_unique_at`).

Invariance under arbitrary segment cuts is a corollary of completeness: splitting spans at interior bytes preserves `denote`, so the canonicalizations are equal — for any number of simultaneous cuts, rather than one cut at a time.

Entry into the verified core is `canonicalize_boundaries`, which runtime-validates the boundary shape (at least two boundaries, the first zero, strictly increasing, one label per gap) and returns `Option`, so no caller carries an unchecked proof obligation. CI verifies these proofs alongside the existing `duvet-coverage` proofs.

*Trusted base:*

The proofs cover canonicalization, not the comparison as a whole. What they rest on:

- **Verus, `vstd`, and Z3**, at the release pinned in CI.
- **`denote` is the definition of "same coverage."** Whether per-byte annotation-set labels are the right notion of equivalence for report comparison is a review judgment, not a theorem.
- **Label interning is unverified glue.** `render_canonical_line` interns each window's annotation set to an integer before calling the verified core; the proofs treat labels as opaque and assume distinct sets get distinct labels. Interning is a `BTreeMap` keyed on the sorted annotation list, and labels never reach the output (the emitted JSON carries annotation lists), but injectivity itself is not proven.
- **Per-window coverage is unverified glue.** Deciding which overlays cover a window (`overlay_start <= start && end <= overlay_end`) happens outside the `verus!` block.
- **The property tests stand in for both.** `canonical_line_is_invariant_under_segmentation_refinement` and `canonical_line_is_invariant_under_overlapping_overlay_refinement` drive the whole path through `render_canonical_line`, comparing two different segmentations — including overlapping, nested, gapped, and reordered overlays — of the same per-byte coverage. They test *invariance*, not content correctness: a bug that renders the same wrong output for both inputs passes. They have teeth against the mutations they target — a "first covering overlay wins" mutation is caught within a few RNG inputs — but that is the shape of the guarantee.
- **The exported contract is one-directional.** `canonicalize_boundaries`'s `ensures` constrains only the `Some` case; an implementation that always returned `None` would verify. `None` maps to a conversion error and is unreachable given how the boundary set is built, but "it returns `Some`" is a test-backed claim, not a proven one.
- **Span ends inside the line text are an invariant by construction, not a check.** Rendering slices `&line.text[span.start..span.end]`; overlays are built from v1 segments whose ends are bounded by the text length, so an out-of-range or non-char-boundary end would panic rather than error. The property tests are ASCII; a Unicode fixture test covers char boundaries.
- **The rest of the converter is unverified** — reparsing, ID assignment, the refs table, status recomputation — and is covered by the integration harness round-tripping every fixture against a directly generated v1 report.

*Known lossy behavior:*

- URL-backed specifications emit the canonical v2 URL, including `.txt` for RFCs.
- Requirement comments are reconstructed from sorted origin ranges.
- Canonical section names replace authored aliases.
- Multiple v2 issue links produce a warning and no v1 issue link unless `--issue-link` is supplied.
- The v1 global `blob_link` is omitted in favor of effective per-annotation links.

The integration harness converts and semantically validates every successful fixture after generating direct v1 and v2 reports. Focused tests cover Markdown, IETF page breaks and aliases, empty quotes, disjoint and Unicode ranges, every annotation type, malformed references and ranges, parser drift, extension buckets, status-table ordering, issue-link override behavior, multi-link behavior, and the hidden-but-invocable CLI contract.

HTML generation is intentionally unchanged in this phase. A follow-up can invoke this conversion internally from merged HTML generation without exposing v1 JSON as a user-facing format.

*Testing:*

- `cargo xtask checks`
- `cargo xtask test --integration`
- `cargo test -p duvet`
- `make -C duvet/www && cargo verus build -p duvet-coverage -p duvet --features verify` (22 verified, 0 errors)


duvet/src/convert.rs
```diff
@@ -0,0 +1,51 @@
...
+
+    /// Compare the converted report with a directly generated v1 report.
+    #[clap(long)]
+    validate_against: Option<Path>,
+}
+
+impl Convert {
+    pub async fn exec(&self) -> Result {
+        let input = json_v2::read_report_v2(self.input.as_ref())?;
+        let compare_issue_link = self.issue_link.is_some() || input.issue_links.len() <= 1;
```
ajrudzitis: **Likely bug: `--issue-link` override forces an issue-link comparison it should suppress.**

```rust
let compare_issue_link = self.issue_link.is_some() || input.issue_links.len() <= 1;
```

When an override is supplied, `compare_issue_link` becomes `true`, so `validate_semantics` compares the converted report's `issue_link` (which is now the override value) against the direct report's own link (from config). If a user overrides to a *different* value while also passing `--validate-against`, they get a spurious `$/issue_link` mismatch failure.

The intent seems inverted: an explicit override means "I'm deliberately changing this link," which should *exclude* it from the comparison. Suggest:

```rust
let compare_issue_link = self.issue_link.is_none() && input.issue_links.len() <= 1;
```

This path is currently untested (the integration harness never passes `--issue-link`), so it's low-severity, but the logic reads as a bug.

duvet/src/report/json_v2_to_v1.rs
```diff
@@ -0,0 +1,1728 @@
...
+    }
+    if !report.annotations.extensions.is_empty() {
+        return Err(duvet_core::error!(
+            "cannot convert report with non-empty annotation extension buckets"
+        ));
+    }
+    Ok(())
+}
+
+fn parse_specifications(report: &ReportV2) -> crate::Result<BTreeMap<String, ParsedSpecification>> {
```
ajrudzitis, line 183: **Design point worth confirming: conversion reparses each spec with the live parser and hard-errors on any drift.**

The v2 schema deliberately stores full spec contents plus section byte-ranges/titles as annotations so consumers can reconstruct *without* reparsing. But the converter reparses each `InlineSource` with the current IETF/Markdown parser, uses the reparsed sections for line structure, and treats the stored section annotations only as a drift guard (`validate_sections` requires a 1:1 match, else a "parser drift" error).

Reparsing is defensible — v2 dropped pre-computed line segmentation, and reproducing v1's page-break-stripped segmentation genuinely needs the parser. But the consequence is that a *merged* report combining packages generated by different duvet versions (or converted by a newer duvet than generated it) can hard-fail conversion even though the byte data is intact — which is exactly the multi-package scenario this plan targets.

Not necessarily wrong, but worth (a) confirming fail-hard is the desired behavior for merged inputs vs. trusting stored section metadata, and (b) surfacing this more prominently as an operational limitation in the CLI help / plan docs.

duvet/src/report/json_v2_to_v1.rs
```diff
@@ -0,0 +1,1728 @@
...
+
+        for section in parsed_spec.specification.sorted_sections() {
+            let section_extent = section_range(section);
+            let mut section_requirements = BTreeSet::new();
+            let mut lines = Vec::new();
+            for line in &section.lines {
+                let Line::Str(line) = line else {
+                    continue;
+                };
+                for _ in line.line_range() {
```
ajrudzitis, line 617: **Readability: obscure `for _ in line.line_range()`.**

This mirrors `json.rs`'s `for lineno in line.line_range()`, but here `segment_line` works off byte ranges and ignores the line number, so if a `Line::Str` slice ever spanned more than one source line this would emit the same segmented line N times. It's correct only because each slice is exactly one line in practice. A comment noting that invariant (or a `debug_assert` that `line_range().len() == 1`) would prevent a subtle future regression.

duvet/src/report/json_v2_to_v1.rs
```diff
@@ -0,0 +1,1728 @@
...
+                    lines.push(segment_line(
+                        line.as_ref(),
+                        line.range(),
+                        references,
+                        &mut section_requirements,
+                    )?);
+                }
+            }
+            for reference in references {
```
ajrudzitis: **Minor: redundant section-requirement population.**

SPEC requirements get inserted into a section's set both via `segment_line`'s `requirements` argument (the `requirements.insert(...)` inside the segmentation loop) and again via this explicit extent loop. It's harmless since it's a `BTreeSet`, but the double path is confusing — one source of truth would read cleaner.

ajrudzitis (commented):
Review of the v2→v1 conversion. Overall this is well-built and clearly serves the merge goal — wiring `duvet convert --validate-against` into the integration harness so every fixture (including real IETF RFCs) is round-trip-validated against a directly-generated v1 report is the right call and gives strong confidence in the 1700-line module. Error handling is thorough and the lossy fields are honestly documented.

A few findings below — one likely bug, one design point worth confirming, and some minor readability/efficiency cleanups. None of the minor items block.

Tests, clippy, and the snapshot all pass locally.

duvet/src/report/json_v2_to_v1.rs
```diff
@@ -0,0 +1,1728 @@
...
+            )
+        })?;
+        segments.push((annotation_ids, status_id, segment_text.to_string()));
+        start = end;
+    }
+
+    Ok(LineV1::Segments(segments))
+}
+
+fn all_ref_statuses() -> Vec<RefStatusV1> {
```
ajrudzitis, line 718: **Minor: hand-rolled 256-entry ref table duplicates domain knowledge.**

`all_ref_statuses` / `ref_status_id` re-encode the bit layout that `json.rs::RefStatus::id()` defines independently. This is *self-consistent* (the converter's table and its emitted `status_id`s agree, and the frontend reads booleans rather than the numeric id, so correctness doesn't depend on matching `json.rs`) — so it's not a correctness risk, just duplicated knowledge with only a partial pinning test. A one-line comment ("layout need not match json.rs, only be internally consistent") would save the next reader the investigation.

duvet/src/report/json_v2_to_v1.rs
```diff
@@ -0,0 +1,1728 @@
...
+            "{} range {}..{} is not on UTF-8 boundaries",
+            context,
+            range.start,
+            range.end
+        ));
+    }
+    Ok(())
+}
+
+fn validate_source_ref(contents: &str, start: usize, end: usize, context: &str) -> crate::Result {
```
ajrudzitis, line 854: **Minor efficiency: `validate_source_ref` clones the entire spec contents on every call.**

It builds a throwaway `SourceFile` from the full contents just to bounds/UTF-8-check a range. For large RFCs this is a needless full-file copy — a direct `contents.get(start..end)` plus a length check would avoid the allocation.

duvet/src/report/json_v2_to_v1.rs
```diff
@@ -0,0 +1,1728 @@
...
+        let origin = vec![ByteRange { start: 0, end: 10 }];
+        let covered = normalize_ranges(&[
+            ByteRange { start: 1, end: 4 },
+            ByteRange { start: 3, end: 8 },
+        ]);
+        assert_eq!(difference_len(&origin, &covered), 3);
+    }
+
+    #[test]
+    fn converts_disjoint_unicode_ranges_and_all_annotation_types() {
```
ajrudzitis, line 1749: **Test coverage gap: IETF conversion only covered by integration tests.**

The unit tests here exercise only Markdown. IETF conversion — the more complex parser, with page-break disjoint ranges and section-alias handling — is covered only by the integration harness. Given how much of `canonicalize` exists specifically for IETF alias/canonical-URL resolution, a focused IETF unit fixture would give faster, more localized feedback than a full integration run.

duvet/src/convert.rs
ajrudzitis: Addressed in 05d3459. An explicit override now suppresses issue-link comparison, while zero/one unmodified input links remain strict. Added a policy unit test and verified the real CLI path with `--issue-link ... --validate-against`.

duvet/src/report/json_v2_to_v1.rs
ajrudzitis: Addressed with a comment documenting that the order intentionally mirrors `json.rs::RefStatus::id`. Internal consistency would render correctly, but exact layout parity is useful for a v1 compatibility converter, so the existing partial pinning test remains alongside that explicit contract.

duvet/src/report/json_v2_to_v1.rs
ajrudzitis: Addressed in 05d3459 by removing the extent-based second path and retaining segmentation as the single source of section requirements. This also fixed a subtle empty-quote discrepancy: title-only SPEC references are top-level requirements but are not encountered in rendered section lines by the legacy writer. Added a regression test.

duvet/src/report/json_v2_to_v1.rs
ajrudzitis: Confirmed: fail-hard is intentional. Reproducing the legacy line stream requires the live parser, and trusting metadata after parser drift could silently produce a semantically different v1 report. I added the operational limitation to `duvet convert --help` and made it explicit in the PR description: parser-incompatible reports must use a compatible Duvet version.

duvet/src/report/json_v2_to_v1.rs
ajrudzitis: Addressed with an explanatory comment. The repetition is deliberate compatibility behavior: the streaming v1 writer emits the complete parser slice once for every source line it spans. I did not add a single-line assertion because that would reject an input the legacy writer can currently serialize.

duvet/src/report/json_v2_to_v1.rs
ajrudzitis: Addressed in 05d3459. `validate_source_ref` now checks length/order and uses `str::get` directly, avoiding a full specification clone and throwaway `SourceFile` for each source reference.

duvet/src/report/json_v2_to_v1.rs
ajrudzitis: Addressed in 05d3459 with a self-contained IETF unit fixture. It exercises a page break and RFC header removal, disjoint ranges, canonical `.txt` identity, authored `.html` URL normalization, and the `1` to `section-1` alias. The full integration RFC suites remain as broader coverage.

duvet/src/report/json_v2_to_v1.rs
```diff
@@ -0,0 +1,1878 @@
...
+                .collect(),
+            statuses,
+            refs: all_ref_statuses(),
+        },
+        warning,
+    ))
+}
+
+fn validate_top_level(report: &ReportV2) -> crate::Result {
+    if report.version != "2.0" {
```
seebees, line 164: I get it, but would you prefer some kind of simver comparison? Maybe we ship with this and update later?

duvet/src/report/json_v2_to_v1.rs
```diff
@@ -0,0 +1,1878 @@
...
+    let mut level = 0;
+    for reference in active {
+        bits |= match reference.anno_type.as_str() {
+            "TODO" => 1,
+            "EXCEPTION" => 2,
+            "TEST" => 4,
+            "IMPLICATION" => 8,
+            "CITATION" => 16,
+            "SPEC" => 32,
+            _ => 0,
```
seebees: nit: This means that any other string contributes noting silently. I think all these strings are internally managed, and an enum might be overkill.... just a completeness edge

duvet/src/report/json_v2_to_v1.rs
```diff
@@ -0,0 +1,1878 @@
...
+            Some(format!(
+                "input contains {} issue links; v1 supports only one, so none was selected (use --issue-link to choose one)",
+                links.len()
+            )),
+        ),
+    }
+}
+
+/// Compare two v1 reports after removing ordering and integer-ID artifacts.
+pub fn validate_semantics(
```
seebees, line 1010: We should consider putting this in its own module? Or perhaps wrapping it in Verus? I'm not going to block on it, but this whole canonicalize business is complicated. And while you don't publish this in `--help` people will find it and use it. And if/when the reports change things get complicate.

While it would be cool to prove that both v1 and a converted v2 produce the same report, this is crazy hard. Because it involves JS, JSON, etc. However, we could talk about segmentation invariance. The spec is stored as a list of cuts. Where the cut falls is arbitrary. So `["hello", " world"]` and `["hel", "lo world"]` are the same. We would want to prove that the comparator is blind to this cut. We could define this like:

```
denote(segments) = (text, coverage)

where
  text     = the concatenation of every segment's text, in order

  coverage : Position → Set<AnnotationKey>
             for each byte position from 0 up to (but not including) length(text),
             coverage(position) = the set of annotation keys attached to the
                                  segment that contains that byte
```

We would then prove

```
∀ seg₁, seg₂ :  denote(seg₁) = denote(seg₂)  ⟹  canonical(seg₁) = canonical(seg₂)
```

We would also care about the converse. Because this would say that the canonicalizer doesn't erase real differences. (they are a bijection)

```
∀ seg₁, seg₂ :  canonical(seg₁) = canonical(seg₂) ⟹ denote(seg₁) = denote(seg₂)
```

duvet/src/report/json_v2_to_v1.rs
```diff
@@ -0,0 +1,1878 @@
...
+    report: &ReportV2,
+    stable_to_integer: &HashMap<String, usize>,
+    parsed: &BTreeMap<String, ParsedSpecification>,
+) -> crate::Result<BTreeMap<usize, RequirementStatusV1>> {
+    let mut statuses = BTreeMap::new();
+
+    for (requirement_id, requirement) in &report.annotations.requirement {
+        let spec = parsed
+            .get(&requirement.origin.src)
+            .expect("requirements were validated while building annotations");
```
seebees: You compute this fact twice. This expect is an invariant in prose. I think it would be better to have the build annotations output be the status input. Something like this. Possession of a ResolvedRequirement is the evidence that resolution succeeded.

```
struct ResolvedRequirement<'a> {
    stable_id: &'a str,
    requirement: &'a RequirementAnnotation,
    spec: &'a ParsedSpecification,   // resolution already happened; this field is the proof
}
```

seebees (commented):
None of these is strictly blocking.

duvet/src/report/json_v2_to_v1.rs
ajrudzitis: I kept the exact comparison and documented the policy in fce4fef. This field is a wire-schema revision, not package SemVer, and the reader/merger also require the exact revision. Even if a future `2.1` is intended to be additive, I would rather opt the converter into it after checking the new schema than silently accept it. Shipping exact matching now and revisiting compatibility when we have the first new revision seems like the safer policy.

duvet/src/report/json_v2_to_v1.rs
ajrudzitis: Addressed in c3aed98. `RangeReference` now carries `ConvertedAnnotationType` and `AnnotationLevel` rather than strings. The status-bit match is exhaustive over `Spec` plus the typed v2 annotation variants, so an unexpected string can no longer silently contribute zero.

duvet/src/report/json_v2_to_v1.rs
ajrudzitis: Addressed across a2828b7, bbc41f5, c31eb7c, and 843ded1. The canonicalizer now lives in `duvet/src/report/canonical.rs` inside a `verus!` block, and CI verifies it with the rest of `duvet`.

The model uses contiguous byte spans whose label interns the annotation set; `denote` expands those spans to the per-byte label sequence. Verus proves that the executable canonicalizer implements the spec and preserves that denotation, and that equal canonical outputs imply equal denotations, so it cannot erase a real coverage difference. `lemma_split_invariance` proves the cut-invariance direction directly for an arbitrary split, prefix, and suffix; repeated splits cover the segmentation differences at issue. Text remains part of the surrounding canonical report comparison, while this proof isolates the arbitrary-cut/coverage property.

This does not claim a formal proof of the entire v1-to-v2 conversion, only the tractable canonicalization property you suggested.

duvet/src/report/json_v2_to_v1.rs
ajrudzitis: Addressed in c09069c using this shape. `build_annotations` now returns `ResolvedRequirement` values carrying the stable ID, requirement, and resolved specification. `build_statuses` consumes those values directly, so it no longer repeats the lookup or relies on an `expect` to restate the invariant.

duvet/src/report/canonical.rs
```diff
@@ -0,0 +1,469 @@
...
+    requires
+        spans_well_formed(left),
+        spans_well_formed(right),
+        canonicalize_spec(left) == canonicalize_spec(right),
+    ensures
+        denote(left) == denote(right),
+{
+    lemma_canonicalize_preserves_denotation(left);
+    lemma_canonicalize_preserves_denotation(right);
+}
```
seebees, line 575: Following up on the segmentation-invariance thread: what landed proves one of the two directions I asked about, and it's worth being precise about which one.

This lemma is the **soundness** direction: `canonical(l) == canonical(r) ⟹ denote(l) == denote(r)`. A validation *pass* can be trusted — canonicalization cannot mask a real coverage difference. For a validator that's the safety-critical direction, so this is genuinely good.

The **completeness** direction — `denote(l) == denote(r) ⟹ canonicalize_spec(l) == canonicalize_spec(r)`, which is what makes "arbitrary segment cuts do not create false mismatches" a theorem rather than an intention — is not proven. Today it rests on `lemma_split_invariance` plus the informal argument that repeated splits cover the rest. That inductive closure over repeated splits is exactly the kind of step the solver should check for us.

The good news: completeness is tractable from what's already here, via canonical-form uniqueness. Three lemmas:

```rust
/// No two adjacent spans share a label (maximal runs).
pub open spec fn coalesced(spans: Seq<SegmentSpan>) -> bool {
    forall|i: int| 0 < i < spans.len() ==>
        (#[trigger] spans[i - 1]).label != spans[i].label
}

/// (a) Canonicalization produces coalesced output.
///     Easy induction over canonicalize_prefix: push_span only appends
///     when labels differ.
proof fn lemma_canonicalize_coalesced(input: Seq<SegmentSpan>)
    requires spans_well_formed(input)
    ensures coalesced(canonicalize_spec(input));

/// (b) Uniqueness: a well-formed, coalesced span list is determined
///     by its denotation. This is the real work.
proof fn lemma_coalesced_unique(l: Seq<SegmentSpan>, r: Seq<SegmentSpan>)
    requires spans_well_formed(l), spans_well_formed(r),
             coalesced(l), coalesced(r), denote(l) == denote(r)
    ensures l == r;

/// (c) Completeness — the advertised property, now a short corollary
///     of (a), (b), and lemma_canonicalize_preserves_denotation.
pub proof fn lemma_canonicalize_complete(l: Seq<SegmentSpan>, r: Seq<SegmentSpan>)
    requires spans_well_formed(l), spans_well_formed(r), denote(l) == denote(r)
    ensures canonicalize_spec(l) == canonicalize_spec(r);
```

Proof sketch for (b): both lists are empty iff the denotation is empty (well-formed spans are non-empty). Otherwise both heads start at byte 0, so `l[0].label == r[0].label` from `denote[0]`. The heads must share an endpoint: if `l[0].end < r[0].end`, contiguity puts `l[1]` at byte `l[0].end` with `l[1].label != l[0].label` (coalesced), but `denote` at that byte is `r[0].label == l[0].label` — contradiction. Equal heads, recurse on tails. One engineering note: the tails don't start at byte 0, so either generalize `spans_well_formed` with a start-offset parameter for the induction, or carry an explicit offset ghost. Effort is comparable to `lemma_canonicalize_prefix_preserves_denotation` above.

With (c) in place the claim in the PR description is fully backed — and, per my comment on `lemma_split_invariance` below, you also get to delete code.

duvet/src/report/canonical.rs
```diff
@@ -0,0 +1,469 @@
...
+        right.end == whole.end,
+        left.label == whole.label,
+        right.label == whole.label,
+    ensures
+        finish_spans(push_span(output, whole), suffix)
+            == finish_spans(push_span(push_span(output, left), right), suffix),
+{
+    reveal(push_span);
+    assert(push_span(output, whole) == push_span(push_span(output, left), right));
+}
```
seebees: Proof-engineering point on this one: the lemma is stated over `finish_spans`, a second recursive formulation introduced only here, and nothing in the file connects `finish_spans` back to `canonicalize_spec` or to the executable `canonicalize_spans`. So the theorem "splitting a span doesn't change canonicalization" is proven about a parallel definition, and its relevance to the deployed function rides on an unproven bridge:

```rust
// Missing: the bridge that makes lemma_split_invariance apply to the real pipeline
proof fn lemma_finish_spans_agrees(input: Seq<SegmentSpan>, k: int)
    requires 0 <= k <= input.len()
    ensures canonicalize_spec(input)
        == finish_spans(canonicalize_prefix(input, k), input.skip(k));
```

To be clear: I believe the two formulations agree and there's no bug today — the `push_span` commutation at the heart of this lemma is a genuine one-step fact. But "the proof is about the code that runs" is what keeps this trustworthy as the code evolves, and right now that link is by inspection only.

Two ways to close it, either is fine:

1. Add the bridge lemma above (straightforward induction on `k`).
2. Preferred: prove completeness per https://github.com/awslabs/duvet/pull/235#discussion_r3693340728. Split invariance then becomes a corollary of the easy fact "splitting one span preserves `denote`" — and both `finish_spans` and this lemma can be deleted outright. Fewer parallel definitions, strictly stronger theorem.

Leaving the choice to you.

duvet/src/report/json_v2_to_v1.rs
```diff
@@ -0,0 +1,1993 @@
...
+            *label
+        } else {
+            let label = annotations_by_label.len();
+            annotation_labels.insert(annotations.clone(), label);
+            annotations_by_label.push(annotations);
+            label
+        };
+        spans.push(SegmentSpan { start, end, label });
+    }
+    let spans = canonicalize_spans(&spans);
```
seebees: `canonicalize_spans` requires `spans_well_formed(input@)`, and because this call site sits outside the `verus!` block, that precondition is an **unchecked assumption** here — Verus never sees this call. Reading the construction, it does hold: `boundaries` is a `BTreeSet` seeded with `{0, text.len()}`, so `windows(2)` yields strictly-increasing contiguous spans starting at 0. But that argument lives in review comments rather than the solver, and it's the single hypothesis every guarantee in canonical.rs hangs from.

Cheap upgrade: move the windows-to-spans step inside `verus!` so the obligation is discharged rather than assumed —

```rust
/// boundaries: strictly increasing, first element 0; labels: one per gap.
pub fn spans_from_boundaries(boundaries: &[usize], labels: &[usize])
    -> (out: Vec<SegmentSpan>)
    requires
        boundaries@.len() >= 2,
        boundaries@[0] == 0,
        forall|i: int| 0 < i < boundaries@.len()
            ==> boundaries@[i - 1] < boundaries@[i],
        labels@.len() + 1 == boundaries@.len(),
    ensures spans_well_formed(out@),
```

"Sorted, distinct, starts at 0" is trivially true of a `BTreeSet` iteration and much easier to audit (or even runtime-check) than the three-clause span invariant.

Second, orthogonal suggestion: a property test that exercises the direction the proofs don't yet cover, end to end through this function — generate random overlays, derive two different refinements of the same per-byte coverage (split spans at random interior bytes, vary which boundaries exist), and assert `render_canonical_line` produces identical JSON for both. That tests completeness (https://github.com/awslabs/duvet/pull/235#discussion_r3693340728) *and* the unverified glue the proof can't see — label-interning injectivity, the containment-as-overlap window computation — in one shot. Until lemma (c) lands, this test is the check standing in for it.

seebees (changes requested):
The Verus work here is genuinely well-engineered — the executable carries `ensures output@ == canonicalize_spec(input@)`, so implementation-matches-spec is the strong form, and the denotation-preservation proof is clean. Requesting changes on one substantive point and two smaller ones:

1. **Soundness is proven; completeness is not.** `lemma_canonical_equality_implies_denotation` gives canonical equality ⟹ equal coverage — canonicalization cannot mask a real difference, which is the right safety direction for a validator. But the other direction I asked for — equal coverage ⟹ canonical equality, the one that makes "arbitrary cuts don't create false mismatches" a theorem — isn't proven; it rests on `lemma_split_invariance` plus prose. Concrete three-lemma plan: https://github.com/awslabs/duvet/pull/235#discussion_r3693340728

2. **`lemma_split_invariance` is proven about `finish_spans`, which is never connected to `canonicalize_spec`** or the executable. Real lemma, unproven bridge. Options (one of which subsumes it and deletes code): https://github.com/awslabs/duvet/pull/235#discussion_r3693342585

3. **`spans_well_formed` is discharged by construction in unverified code** at the sole call site. Suggested verified constructor plus a property test that covers the completeness gap and the interning glue through the executable path: https://github.com/awslabs/duvet/pull/235#discussion_r3693344755

One more ask for the PR description: name the trusted base explicitly — the precondition-by-construction at the call site, label-interning injectivity, and the per-window coverage computation are unverified glue exercised by the integration harness. That boundary is fair and well-designed; it should just be written down next to the proof claims.

duvet/src/report/canonical.rs
ajrudzitis: Completeness landed in 1299a7a, following your three-lemma plan closely enough that the names line up.

**(a) Canonical output is coalesced.** Rather than a standalone lemma, this folded into the existing prefix induction: `lemma_push_preserves_coalesced` is the one-step fact, and `lemma_canonicalize_prefix_preserves_denotation` carries `coalesced` alongside well-formedness and denotation, so `lemma_canonicalize_preserves_denotation` now ensures all three:

```rust
pub proof fn lemma_canonicalize_preserves_denotation(input: Seq<SegmentSpan>)
    requires spans_well_formed(input),
    ensures
        spans_well_formed(canonicalize_spec(input)),
        coalesced(canonicalize_spec(input)),
        denote(canonicalize_spec(input)) == denote(input),
```

**(b) Uniqueness.** `lemma_coalesced_denotation_unique_at(left, right, offset)`. I took the offset-parameter option you flagged: `spans_well_formed_at(spans, offset)` is the generalized invariant and `spans_well_formed(spans) == spans_well_formed_at(spans, 0)`, so the induction can recurse into tails that no longer start at byte 0. The proof follows your sketch — both empty iff the denotation is empty (well-formed spans are non-empty); otherwise both heads start at `offset` so `denote[0]` forces equal labels; if `left[0].end < right[0].end` then contiguity puts `left[1]` at that byte with a different label by `coalesced`, contradicting `denote` there; equal heads, then recurse via `lemma_well_formed_drop_first` / `lemma_coalesced_drop_first`.

One piece the sketch didn't anticipate: `denote` recurses on the *last* span, which is what makes it line up with `push_span`, but the uniqueness induction peels the *front*. Instead of restating `denote`, I added `denote_forward` and proved the two agree (`lemma_denote_forward`, via `lemma_denote_forward_push`), so uniqueness can work front-first while every other lemma stays over the original definition.

**(c) Completeness.** `lemma_canonicalize_complete` is the short corollary you predicted:

```rust
pub proof fn lemma_canonicalize_complete(left: Seq<SegmentSpan>, right: Seq<SegmentSpan>)
    requires spans_well_formed(left), spans_well_formed(right),
             denote(left) == denote(right),
    ensures canonicalize_spec(left) == canonicalize_spec(right),
{
    lemma_canonicalize_preserves_denotation(left);
    lemma_canonicalize_preserves_denotation(right);
    lemma_coalesced_denotation_unique_at(canonicalize_spec(left), canonicalize_spec(right), 0);
}
```

So "arbitrary segment cuts do not create false mismatches" is now a theorem rather than an intention, and the PR description has been rewritten to state both directions instead of the old split-invariance phrasing. And yes — I got to delete code; see the reply on `lemma_split_invariance`.

`cargo verus build -p duvet-coverage -p duvet --features verify` reports 22 verified, 0 errors.

duvet/src/report/canonical.rs
ajrudzitis: Took option 2. `finish_spans` and `lemma_split_invariance` are both deleted as of 1299a7a, so the unproven bridge is gone rather than patched — there is no second recursive formulation left in the file, and every lemma is now stated over `canonicalize_spec`, which the executable carries as `ensures output@ == canonicalize_spec(input@)`.

Split invariance survives as a corollary of completeness, and in a stronger form than the deleted lemma had: splitting a span at an interior byte doesn't change the per-byte label sequence, so `denote` is equal, so `lemma_canonicalize_complete` gives equal canonicalizations. That holds for any number of simultaneous cuts at once, not one cut at a time with an informal induction over repetitions — which was the part that needed the solver.

duvet/src/report/json_v2_to_v1.rs
ajrudzitis: Both suggestions are in.

**Verified constructor.** Rather than a `requires`-based `spans_from_boundaries` — which would move the obligation but still leave it discharged by inspection at an unverified call site — I made the boundary shape a *runtime* check, so there is no unchecked assumption at the boundary at all. `canonicalize_boundaries` (1299a7a, contract exported in 2d76b98) returns `Option`, with `None` for fewer than two boundaries, a first boundary other than zero, a label count that doesn't match, or a non-increasing pair:

```rust
pub fn canonicalize_boundaries(boundaries: &[usize], labels: &[usize])
    -> (output: Option<Vec<SegmentSpan>>)
    ensures
        output is Some ==> {
            &&& output->Some_0@ == canonicalize_spec(induced_spans(boundaries@, labels@))
            &&& spans_well_formed(output->Some_0@)
            &&& coalesced(output->Some_0@)
            &&& denote(output->Some_0@) == denote(induced_spans(boundaries@, labels@))
        },
```

`induced_spans(boundaries, labels)` names exactly the span list a boundary/label pair denotes (span `j` runs `boundaries[j] .. boundaries[j + 1]` carrying `labels[j]`), so the contract is stated in the caller's vocabulary: the guarantee is about *your* boundaries, not about a span invariant you have to re-derive. `render_canonical_line` maps `None` to a conversion error with a comment noting it is defensively unreachable given the `BTreeSet` construction — the point being that unreachability is now a remark about an error path, not the hypothesis every guarantee in `canonical.rs` hangs from.

**Property test.** `canonical_line_is_invariant_under_overlapping_overlay_refinement` (5fbd90c), alongside the existing refinement property. It draws up to 11 overlays independently over a random line, so they overlap, nest, and leave gaps — which is what `merge_canonical_specification` actually produces when duplicate RFC-target specifications get their overlays appended onto one line. It then builds a different segmentation of the same per-byte union by splitting overlays at interior bytes and duplicating some, and asserts `render_canonical_line` emits byte-identical JSON for both. It also reverses the refined list, since the union is insensitive to arrival order as well as to where the cuts fall.

That last part earned its place: I mutation-tested the property. A "first covering overlay wins" mutation (a `break` after the annotation union) passes the entire 291-test unit suite, *including* both property tests, when the reversal isn't there; with it, the mutant fails within a couple of RNG inputs. A mutation to the containment check is caught immediately either way.

On what these tests do and don't buy: they exercise exactly the unverified glue you named — label-interning injectivity and the containment-as-overlap window computation — end to end through `render_canonical_line`. But they compare two *renderings* against each other, so they establish invariance, not content correctness: a bug that renders the same wrong output for both segmentations passes. I've written that out, along with the rest of the trusted base, in the PR description.

ajrudzitis:
Thanks — all four items from https://github.com/awslabs/duvet/pull/235#pullrequestreview-4832175566 are addressed. Details are in the three threads; summarizing here since the fourth ask has no thread of its own.

1. **Completeness** — proven in 1299a7a, following your three-lemma plan; `lemma_canonicalize_complete` is the corollary. [Thread](https://github.com/awslabs/duvet/pull/235#discussion_r3732855014)
2. **`lemma_split_invariance` / `finish_spans`** — took the option that deletes code: both are gone, and split invariance is now a corollary of completeness covering arbitrarily many simultaneous cuts. [Thread](https://github.com/awslabs/duvet/pull/235#discussion_r3732855668)
3. **`spans_well_formed` by construction** — `canonicalize_boundaries` runtime-validates the boundary shape and returns `Option`, so nothing is assumed at the call site, and its `ensures` exports the contract in terms of the caller's boundaries. Plus the property test over overlapping/nested/gapped/reordered overlays. [Thread](https://github.com/awslabs/duvet/pull/235#discussion_r3732855828)
4. **Trusted base in the description** — done, as its own section. It names `denote` as the chosen equivalence relation, label-interning injectivity and the per-window containment computation as unverified glue, the property tests as invariance-only stand-ins for those (they compare two renderings, so identical-but-wrong output passes), the one-directional `ensures` on `canonicalize_boundaries`, and the by-construction span-end/char-boundary invariant behind the rendering slice. The old "invariant under splitting a span at an arbitrary interior byte" line is replaced by an explicit statement of both proof directions.

Rebased on main; CI is green, and `cargo verus build -p duvet-coverage -p duvet --features verify` reports 22 verified, 0 errors. Ready for another look when you have time.

seebees (dismissed):
**Approving.** One non-blocking item: 

**We should remove `vstd` from the published dependency graph.** Right now it's an unconditional dependency of `duvet`, which means: (a) the exact pre-release pin `=0.0.0-2026-05-24-0157` lands in a published library — pre-release exact pins never unify, so any downstream lib-consumer carrying its own vstd pin gets an unresolvable graph, and Verus-adjacent tooling is exactly who consumes duvet as a library; (b) every `cargo install duvet` compiles vstd even though every vstd path in `canonical.rs` is ghost and erased; (c) downstream SBOMs and dependency audits grow by the verus crates for zero shipped code. It also diverges from the principle duvet-coverage's own Cargo.toml states: "nothing verus-shaped reaches the published artifact."

**I verified the duvet-coverage pattern applies directly.** Three edits:

1. `duvet/Cargo.toml`: `verus_builtin_macros` unconditional; `vstd` and `verus_builtin` optional behind `verify = ["dep:vstd", "dep:verus_builtin", "duvet-coverage/verify"]`
2. `canonical.rs`: `use verus_builtin_macros::verus;` plus `#[cfg(feature = "verify")] use vstd::{assert_seqs_equal, prelude::*};`
3. CI command unchanged — `cargo verus build -p duvet-coverage -p duvet --features verify` works as-is

If you have a case for keeping vstd unconditional — simpler lockstep maintenance, binary-dominant usage — I'm fine; I just want it to be an explicit decision rather than a default. :)

seebees (approved):
LGTM!

PR #227, opened 2026-05-28
feat: duvet query — interactive traceability checks with Verus-verified coverage model

seebees:
## Summary
  Adds `duvet query`, a development-time companion to `duvet report`
  that answers focused traceability questions during the inner-loop
  ("have I annotated this requirement?", "does my test execute the
  implementation it claims to test?"). Adds a new internal crate
  `duvet-coverage` that contains the formally-verified two-phase
  coverage model used by the coverage check.

  ## Motivation

  `duvet report` generates the full traceability artifact for CI but is
  not optimized for the development inner loop. Engineers iterating on
  a single requirement need fast, focused, filterable feedback —

  The coverage check correlates test annotations with executed
  implementation annotations using a coverage report. For Java sources
  (via JaCoCo XML), it uses a tree-sitter classifier and a verified
  two-phase algorithm that handles source constructs invisible to
  bytecode coverage: method declarations, interface bodies, fields
  without initializers, etc. For other languages it uses a verified
  degraded model that reads coverage directly at the annotation's
  target line.

  ## What's new

  ### Command
  - `duvet query --check {implementation,test,coverage,executed-coverage,duplicates}`
    — five composable checks
  - `--section`, `--quote` filters (repeatable, comma-separated)
  - `--verbose` to show passing checks and the coverage-model
    path-selection summary

  ### Crates
  - New `duvet-coverage` (workspace-internal, `publish = false`).
    Pure-function coverage model:
    - `types` — `LineProperty`, `LineClass`, `Scope`, `ExecutionStatus`,
      `CoverageReport`
    - `scopes::build_scope_tree` — Phase 1 input
    - `target_resolution::annotation_target` — internal Phase 1
    - `execution_propagation::execution_set` — internal Phase 2
    - `annotation_execution::is_annotation_executed` — public entry
    - `predicates`, `proofs` — Verus-only correctness machinery
  - `duvet` gains `query`, `query/checks/coverage`,
    `query/parsers/jacoco`, `query/classify/java`,
    `query/classify/mod`

  ### Verification
  - Six correctness properties proven with Verus over the algorithms
    in `duvet-coverage`. The proofs live alongside the algorithms in
    `duvet-coverage/src/proofs.rs`.
  - Per-property runtime smoke tests give concrete-input documentation
    of each property; the proof gives universal correctness, the tests
    give a runnable example.
  - New CI job (`verify`) downloads the pinned Verus release, installs
    its required Rust toolchain, and runs `cargo verus build -p duvet-coverage`
[...]

PR #236, opened 2026-07-22
ci(deny): pass --config as a pre-subcommand common option

seebees:
## Summary

Fixes the `dependencies` workflow, which currently fails on every run: `--config` (and `--all-features`) are cargo-deny *common* options and must precede the subcommand. The EmbarkStudios/cargo-deny-action assembles them from the `arguments` input (placed before `command`); appending `--config` after `check` fails with `unexpected argument '--config'`.

## Testing

- The identical change has been running green on the #227 branch (`deny` check passes there).

Split out of #227 so CI is repaired for every PR, not just that one.

### DIFF
--- a/.github/workflows/dependencies.yml
+++ b/.github/workflows/dependencies.yml
@@ -32,4 +32,9 @@ jobs:
 
       - uses: EmbarkStudios/cargo-deny-action@v2
         with:
-          command: check --config .github/config/cargo-deny.toml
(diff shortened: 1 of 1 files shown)


boquan-fang (approved):

ajrudzitis (approved):

seebees:
Superseded by #237 ("ci: make main pass again"), which combines all three main-CI fixes so the repair PR itself can go fully green — no individual fix can, because the three breakages mask each other.

PR #240, opened 2026-07-22
style(report): drop needless borrow in json report id formatting

seebees:
## Summary

`format!("{}", &source.path)` borrows for no reason; format args already take the value by reference. One-character cleanup.

## Testing

- `cargo build -p duvet` passes.

Split out of #227.


seebees:
Folded into #237 ("ci: make main pass again") — this turned out not to be a cosmetic nit: Rust 1.97 clippy's `useless_borrows_in_formatting` lint fires on this line with `-D warnings`, so it is the fourth independent main-CI breakage.

PR #244, opened 2026-07-28
fix(query): pair discharge requires a single witness report

seebees:
## Summary

The `coverage` check folded a test annotation's and an implementation annotation's execution status **independently** across all supplied coverage reports, so no single report was ever required to have seen both. With per-test reports (the workflow that gives coverage correlation its meaning), a test annotation could be discharged by an implementation that a *different* test executed — exactly the vacuous pairing the check exists to catch.

This change requires a single common witness: a pair (test, impl) passes only when at least one report shows the test executed **and** that same report shows the implementation executed.

- `design/query/decisions.md` gains Decision 13 recording the semantics and why the original OR-fold decision's con (scoped to aggregate reports) was over-generalized by the implementation
- `design.md` §5.2 rewritten to state the per-witness intent
- New integration test `query-coverage-cross-report-pair` pins the failing direction (test executed in report A, impl executed only in report B ⇒ FAIL); the existing `query-coverage-multi-report-or` test is byte-identical under the fix (its report contains both sides)

## Testing

- New discriminator test fails as intended (exit 1, diagnostic names the witness that missed the implementation)
- `query-coverage-multi-report-or` snapshot unchanged
- `cargo test -p duvet query::` green
- Verified coverage model (`duvet-coverage`) unchanged — the per-annotation scoring beneath the fold is untouched

## Note

First of a series: a follow-up PR builds a witness abstraction on this semantics to admit prover-based coverage sources (Verus SST elaboration records) alongside runtime reports.

### DIFF
--- a/design/query/decisions.md
+++ b/design/query/decisions.md
@@ -315,6 +315,13 @@ and other test suites that produce separate reports.
 The `--coverage-report` flag accepts multiple paths
 with glob expansion.
 
+Refined by [Decision 13](#decision-13):
+OR semantics remain correct for per-annotation questions
+("was this ever executed?")
+and for false-failure prevention,
+but discharging a test/implementation pair
+requires a single report that witnessed both sides.
+
 ---
 
 ## Decision 7: Executed coverage variant {#decision-7}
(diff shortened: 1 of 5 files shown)


design/query/decisions.md
```diff
@@ -655,3 +662,100 @@ So Option B's intent — "languages without classifiers keep working, no
...
+Quantifier scope is part of the meaning of the check,
+not an implementation detail of the fold.
+
+This decision is the report-file instantiation of the general
+witness model specified in
+[`design/witness/spec.md`](../witness/spec.md)
+(Property W1: same-witness discharge),
+where witnesses may also be prover obligations;
+see [`design/witness/decisions.md`](../witness/decisions.md)
+for that feature's design record.
```
seebees, line 761: This precedes the coming PR about using proof as a form of execution. A witness is the more abstract language for what is today a report.

dougch (approved):
Struggled with some of the language- Claude helped clear things up 👍

Issue #226, opened 2026-05-28
duvet's annotation parser scans Rust string literals, causing accidental traceability


seebees:
## Summary

  duvet's annotation parser treats `//=` and `//#` patterns as real
  annotations regardless of whether they appear in source comments or
  inside Rust string literals. As a result, example annotations placed
  inside doc / help strings — for documentation purposes — are picked
  up as real traceability statements.

  `duvet/src/query.rs` defines `enum CheckType` whose variants carry
  `#[value(help = "...")]` strings that document each `--check` mode.
  Several of those help strings include example annotations of the form:

  ```text
  //= https://www.rfc-editor.org/rfc/rfc2324#section-2.1.1
  //# A coffee pot server MUST accept both the BREW and POST method
  //# equivalently.
```

  When duvet runs against itself, those examples are recorded as real
  annotations targeting RFC 2324 (HTCPCP). Consequences:

  - .duvet/snapshot.txt lists https://www.rfc-editor.org/rfc/rfc2324

    as a tracked specification.

  - The spec text is cached at

    .duvet/specifications/www.rfc-editor.org/rfc/rfc2324.txt (~19 KB)
    and committed to the repository so builds remain reproducible
    without a network fetch.

  - A naive cleanup of the cached file would break the duvet check.

  Why a fix is wanted

  - Documentation should not create permanent build dependencies.
  - The same trap will fire for any project that documents duvet

    a comment explaining "do not delete this" — is technical debt.

  Possible fixes (in increasing scope)

  1. Escape the example annotations in duvet/src/query.rs so the

     parser ignores them. Insert a non-pattern character (e.g. // =
     with a space) or move help text out into doc comments /
     include_str! from a .txt file.

     contents.** When scanning a .rs file, parse just enough of the
     token stream to know whether each line is inside a string literal,
     and if so, do not match annotation patterns there.

  3. **Generalize: the annotation extractor should be source-language

     aware** enough to skip any construct the source language treats as
     data, not as comment (string literals, raw strings, byte strings,
     etc., for Rust; equivalents in other languages).

PR #249, opened 2026-08-04
ci: run duvet's own annotation checks on every build (dogfood)

seebees:
# ci: run duvet's own annotation checks on every build (dogfood)

## What gates

A dedicated `dogfood` job with two named, exit-code-only steps — no output
parsing, no thresholds, no expected-failure lists:

1. **Run duvet on its own annotations** — `duvet report --ci`: the committed
   `.duvet/snapshot.txt` pins every requirement's coverage state; any drift
   fails.
2. **Annotation checks** — `duvet query -c implementation,test,duplicates`:
   every spec requirement carries an implementation annotation, every
   implementation annotation carries a test annotation, and no (type, quote)
   pair has more than one owner.

The job mirrors the shape #247 extends (it adds SST-artifact download and the
witness-query slice to this same job), so #247's rebase is additive.

## Annotation debt fixed (measured on main @ 3d6c3d1)

Before: Implementation FAIL (7 of 21 requirements unannotated, all in
`design/query/coverage-model-spec.md`), Tests FAIL (1 untested), Duplicates
PASS. After: all three PASS. Per-quote, with the convention applied:

| Spec quote | Fix | Convention |
|---|---|---|
| §1.3 `{Annotation}` excludes `Statement`/`Declaration` | impl on `clean_classifications`, test above the stripping assert | citation at implementing site; test at the verifying assert |
| §1.3 classifiers MUST apply a post-processing pass | same file, impl + test | same |
| §1.5 Scope Balance Contract | impl on `scope_imbalance_site` (the enforcement point — the emitting classifiers are unscannable fixture carriers), test in `stray_close_is_flagged_and_tree_collapses` | one owner per (type, quote), pointer comment at the detector |
| Property 11 balance detection | impl on the fn header; **`type=test` on the `ensures` clause** — the clause that verifies is the test's anchor | test annotations for proven properties anchor on the ensures clause (the discharge unit #247's witness machinery roots) |
| §7.4 "MUST be proven with Verus" | citation on the first D-property proof site (`degraded_execution_status`, whose `ensures` carry D1/D2), `type=test` on the runtime demonstration test (migrates to the CI verify step with #247) | meta-requirements anchor on the first proof-bearing item, never on imports |
| §1.5 unbalanced stream MUST NOT be scored / §8.2 | impl citations on the engine's defeated-classification reporting block | spec's own §8.3 note: citations land at the implementing sites in the `duvet` crate |
| §8.2 test obligations | **quote split**: the non-scoring clause is tested by `defeated_classification_is_surfaced_not_scored`; the "reporting each located issue" clause by `escalation_reports_each_located_issue`, over the extracted `collect_defeated_issues` aggregation the loud report iterates | a test annotation quotes only what its asserts verify |
| §8.3 `report` hard error on defeated commitment | `type=exception` with reason — `report` does not consume coverage yet, the conditional is vacuous until the LCOV follow-up | exception for genuinely unimplementable-today text; follow-up: remove the requirement from the spec until the LCOV PR implements it |

## Two items reviewers should look at

1. **`.duvet/config.toml` now enumerates two fixture-clean `duvet/src` files**
   (`query/engine.rs`, `report.rs`) as scanned sources, so the §8 citations can
   live at their implementing sites. `duvet/**` wholesale still waits on #226;
   the config note lists the known fixture carriers.
2. **Two unit tests beyond pure annotation fixes.** The §1.5/§8.2 escalation
   citations create test obligations whose existing verifying tests live only
   in unscannable fixture-carrier files. Retyping the annotations to dodge the
   check would be deciding the property true instead of proving it, so the
   tests carry the obligations honestly: a `DefeatedClassification` entry
   resolves to a located `Unknown` — never scored against the collapsed tree —
   and the escalation aggregation preserves each located issue from every
   defeated file (none dropped, none merged away). The aggregation is extracted
   into `collect_defeated_issues` so the second test executes exactly the code
   the loud report iterates.

The snapshot is regenerated; `duvet report --ci`, `duvet query -c
implementation,test,duplicates`, unit tests, and clippy (`-D warnings`) all
pass.

## Relation to #247

#247 extends this same `dogfood` job with proof-witness coverage gates
(`query -c coverage` over Verus SST logs). This PR is independent of and ahead
of it: it lands the annotation-hygiene floor on main so #247 rebases onto a
tree where the job exists and the three basic checks already gate. The
ensures-clause test placement here is forward-compatible by design: under
#247's discharge-unit model (spec §5.3), that annotation roots the clause-level
unit and is discharged by the verus-sst proof witness with no further move.

### DIFF
--- a/duvet/src/report.rs
+++ b/duvet/src/report.rs
@@ -61,12 +61,8 @@ pub struct Report {
 impl Report {
     //= design/query/coverage-model-spec.md#escalation
     //= type=exception
-    //= reason=`report` does not consume coverage yet: no coverage source
-    //= reason=reaches this path today, so the conditional ("When `report`
-    //= reason=consumes coverage") is vacuous. The hard-error obligation
-    //= reason=lands with the LCOV mixed-coverage follow-up that wires
-    //= reason=coverage into report; until then there is no code site that
-    //= reason=could implement it.
+    //= reason=`report` does not yet consume coverage; the hard-error gate lands with the LCOV follow-up.
+
     //# When `report` consumes coverage, a defeated commitment MUST be a hard
     //# error there, because `report` is the authoritative artifact.
     pub async fn exec(&self) -> Result {
--- a/duvet/src/report.rs
+++ b/duvet/src/report.rs
@@ -59,6 +59,16 @@ pub struct Report {
 }
 
 impl Report {
+    //= design/query/coverage-model-spec.md#escalation
+    //= type=exception
+    //= reason=`report` does not consume coverage yet: no coverage source
+    //= reason=reaches this path today, so the conditional ("When `report`
+    //= reason=consumes coverage") is vacuous. The hard-error obligation
+    //= reason=lands with the LCOV mixed-coverage follow-up that wires
+    //= reason=coverage into report; until then there is no code site that
+    //= reason=could implement it.
+    //# When `report` consumes coverage, a defeated commitment MUST be a hard
+    //# error there, because `report` is the authoritative artifact.
     pub async fn exec(&self) -> Result {
         let config = self.project.config().await?;
         let config = config.as_ref();
--- a/duvet/src/report.rs
+++ b/duvet/src/report.rs
@@ -62,7 +62,6 @@ impl Report {
     //= design/query/coverage-model-spec.md#escalation
     //= type=exception
     //= reason=`report` does not yet consume coverage; the hard-error gate lands with the LCOV follow-up.
-
     //# When `report` consumes coverage, a defeated commitment MUST be a hard
     //# error there, because `report` is the authoritative artifact.
     pub async fn exec(&self) -> Result {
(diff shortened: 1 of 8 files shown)


duvet/src/report.rs
```diff
@@ -59,6 +59,16 @@ pub struct Report {
...
 
 impl Report {
+    //= design/query/coverage-model-spec.md#escalation
+    //= type=exception
+    //= reason=`report` does not consume coverage yet: no coverage source
+    //= reason=reaches this path today, so the conditional ("When `report`
+    //= reason=consumes coverage") is vacuous. The hard-error obligation
+    //= reason=lands with the LCOV mixed-coverage follow-up that wires
+    //= reason=coverage into report; until then there is no code site that
+    //= reason=could implement it.
```
dougch: `duvet report --ci` is failing for me locally with `× "reason" already specified`; claude's fix:

```suggestion
    //= type=exception
    //= reason=`report` does not yet consume coverage; the hard-error gate lands with the LCOV follow-up.

```

dougch (approved):
Point of confusion for me- older versions of duvet fail on these changes...

PR #256, opened 2026-08-07
Duplicates check v2 + duplicate-targets query (design/duplicates spec)

seebees:
## What

Turns `design/duplicates/decisions.md` into a normative spec and implements it:

- **`design/duplicates/spec.md`** — RFC 2119 spec for duplicate policing over the two coincidence axes the decisions doc formalized: claim classes (annotations sharing a section + normalized quote) and target classes (annotations sharing a resolved source position).
- **One `duplicates` check, both axes** (`duvet query --check duplicates`) — a single invocation evaluates every configured predicate over both partitions, mirroring the one `[duplicates]` config table:
  - **Claim axis:** §2.1 same claim + same resolved target fails unconditionally, any form pair (Decision 2) · §2.2 per-form multiplicity caps, default 1, configurable only for `test`/`implementation` (Decisions 1, 3) · §2.3 subsumption fails cap-independently (legacy union-coverage semantics, separated from the cap rule) · §2.4 free-form exclusivity within a claim class (Decision 4 — the one shipped tightening, restorable with one config line).
  - **Target axis (Decisions 8, 12):** fan-in listing in every report (one-line summary by default, full listing under `--verbose`: count descending, per-form breakdown, distinct-section count) · opt-in `count`/`sections` bounds · form-set family — `test`+`implementation` on one target fails by default.
- **Config**: `[duplicates.claims]` / `[duplicates.targets]` (Decisions 9, 11). Verdict is a function of the checked-in tree; the CLI never changes a verdict. `citation` accepted on input, `implementation` emitted (Decision 10). Unknown keys and non-positive bounds are configuration errors. Keys documented in the guide, the `duvet init` template, and `--help`.
- **Defaults settled by the corpus survey** (see comments below): `targets.count` defaults unlimited — the legitimately dense population is real (s2n-quic bears 149 multi-annotation targets); the observed max is 5, so strict teams can write `count = 5` today. Zero external targets mix `test`+`implementation`, validating the default family.
- **Dogfood**: this tree passes its own policy (`targets.count = 1` — every fan-in site dissolved under more targeted placement) and CI gates it with a `Self-check duplicates` step running the branch-built binary, unsliced. The spec is registered in `.duvet/config.toml` with a full traceability chain (implementation + test citations, fixture-pinned).

Target resolution reuses the **verified** `duvet-coverage` forward walk over static classification — no coverage data, no witness layer (Decision 5: independent contracts). A defeated classification fails the run and renders no verdict.

## Spec resolutions beyond the decisions doc (reviewers: these are the judgment calls)

1. **Claim sharing = equality of (section, normalized quote)** — an equivalence relation by construction, discharging the "is mutual coverage transitive?" follow-up.
2. **Subsumption is its own rule (§2.3)**, kept from legacy: a fragment covered by the union of same-form quotes fails regardless of caps. A raised cap admits equal copies only, never fragments (Decision 14's dangerous direction).
3. **Decision 4's coherence config is a form-set family** (`claims.types`), the same primitive Decision 12 gave the target axis.
4. **`spec`-form annotations are excluded from cross-form rules** — every extracted requirement is present as one; a literal reading of Decision 4 would fail every legitimate `exception`.
5. **P-D3 scope**: legacy verdict parity holds at caps 1 with the families fully open, except two named residuals — empty-quote annotations (legacy failed them by accident of an early return) and cross-form same-target stacks (§2.1 is deliberately knobless per Decision 2). §4.4 records the restoration recipe.
6. **Recorded follow-up, not spec'd**: a free form with a strictly broader quote covering a priced claim evades §2.4 (exclusivity keys on equal claims). Fixing it means exclusivity over coverage relations — beyond what Decision 4 decided.

## Testing

- 286 unit tests pass, including policy-semantics tests (default families, fixed free-form caps, `citation` alias round-trip, schema rejection of unknown keys and unpriced caps).
- Integration suite passes in assert mode. 16 new fixtures pin P-D1, P-D2, P-D4 (fixture level), P-D5, P-T1, P-T2, P-S0 (verbosity axis), the Decision 14 restoration paths, the provenance-keyed TOML exclusion, defeated-classification failure, and legacy verdict parity on the exact-pair regression (same FAIL, now decomposed into cap violation + subsumption).
- Corpus runs against s2n-quic, s2n-tls, h3, MPL, and duvet itself (details in comments below) — including real catches on h3 (an `exception`+`todo` contradiction on one RFC 9114 requirement) and on this tree.
- clippy clean, fmt applied.


seebees:
## Corpus run: the checks against real duvet projects

Ran both checks with this branch's binary against s2n-quic, s2n-tls, h3, aws-cryptographic-material-providers-library (MPL), and duvet's own tree (witness-coverage branch).

### Fan-in survey (settles the `targets.count` default — Decision 8's open question)

| corpus | annotations | targets ≥2 | distribution |
|---|---|---|---|
| s2n-quic | 2,425 | 149 | 113×c2, 26×c3, 8×c4, 2×c5 |
| duvet (self) | 356 | 42 | 30×c2, 6×c3, 4×c4, 2×c5 |
| MPL | 1,995 | 26 | 25×c2, 1×c3 |
| h3 | 529 | 21 | 13×c2, 3×c3, 1×c4, 4×c5 |
| s2n-tls | — | 0* | *unreliable: C has no line classifier, and the degraded path can't skip comment lines, so block-comment annotations resolve to distinct comment lines |

**Conclusion:** the legitimately dense population is large; `count = 1` as a default fails every surveyed project on day one. Default stays unlimited (spec updated, no longer provisional). The observed max is 5, so teams wanting bite can write `count = 5` today.

### Default target-form family, validated

Among 238 multi-annotation targets across four reliable corpora, **zero** external targets mix `test`+`implementation` — the default family's prohibition matches real practice. The one hit is in duvet's own tree (`verus_sst/tests.rs:994`, test:2+implementation:1) — a genuine placement bug to fix.

### Duplicates check findings

- **h3** (clean run, real catches): 20 over-cap duplicate sets, 1 subsumed fragment, and 8 exclusivity violations — including an `exception`+`todo` **contradiction** on one RFC 9114 §4.4 requirement ("we don't do this" ∧ "we'll do it later") and two `implementation`+`todo` coexistences (the lifecycle violation Decision 4 predicts: the todo should have been replaced, not joined).
- **duvet itself**: 3 same-claim-same-target stacks (P-D1 violations in `scopes.rs`, `witness.rs`, `verus_sst`) and 4 free-form exclusivity hits (`test`+`implication` sharing claims — the implication-deflation exploit, live in our own tree). Dogfooding burn-down material.
- **s2n-quic / s2n-tls / MPL**: the check aborts on pre-existing quote-location errors (742 / 28 / 76 "Exactly matchable quote not found") before rendering a verdict. Same matcher as main — the legacy check's per-type `try_join_all` short-circuited after the first form's errors (main shows only 4 on s2n-quic), while v2 aggregates all forms. Full visibility, but it means the duplicates check is unusable on corpora whose requirement TOMLs don't exactly match the query matcher — a query-engine matching gap, not a duplicates-rule gap, worth its own issue.

seebees:
## Dogfood: this tree now passes its own checks (6776083)

Running the new checks on this branch's own tree caught 4 free-form exclusivity violations and 3 fan-in sites. Every one was an annotation bug the rules were designed to catch:

- **Composed end-to-end lemmas typed `implication`** (Properties 2/3/4) while being machine-checked Verus proofs — the exact under-claim that deflates a requirement's obligations. Retyped `implementation`.
- **A runtime `#[test]` claiming "These properties MUST be proven with Verus"** — wrong artifact for the claim. It now lives on the CI verify step that runs the prover (`#=`/`#%` source entry, same pattern as the witness branch).
- **D1/D2 implications on the function head** — moved onto their enforcing lines.

### Project settings (each value earned, not defaulted)

```toml
[duplicates.claims]
implementation = 2   # core lemma + composed end-to-end lemma per property

[duplicates.targets]
count = 1            # one target, one annotation — every fan-in site dissolved
```

The `count = 1` outcome is the interesting one: challenged to find a fan-in site that *couldn't* be more targetedly placed, I found none on this tree — even the file-header "MUST be proven with Verus" citation had a sharper home (the CI step that enforces it). The shipped *default* stays unlimited (the corpus survey stands: s2n-quic's 149 dense targets are real), but this project demonstrates the strict posture is livable when placement is honest — and CI now gates it (`Self-check duplicates` step, branch-built binary).

seebees:
## Merged: one `duplicates` check, both axes (80ca5f9)

`duplicate-targets` as a separate check type is removed — it over-read one noun phrase in Decision 8. The model's own symmetry says claim classes and target classes are two partitions of one formal space, and the config was already one `[duplicates]` table with two namespaces; the evaluator now matches. `duvet query --check duplicates` evaluates every configured predicate on both axes, and its report always carries the fan-in listing.

Confirmed and now recorded in spec §4.4: **the historical check remains reachable by configuration** — caps default to 1, bounds default unlimited, and one all-forms family entry per namespace turns the coexistence rules off. P-D3 names the only two residuals: cross-form same-claim-same-target stacks (Decision 2 is deliberately knobless) and the legacy empty-quote accident.

All snapshots regenerated; suite passes in assert mode; the CI self-check step is a single invocation.

seebees:
## Discoverability (fb9a162)

A shipped default that never appears in a config file is invisible — nobody learns the target axis is configurable. Three surfaces now implement Decision 9's discovery loop:

1. **`--verbose` prints the effective policy**, every value, defaults labeled:
```
Effective policy ([duplicates] in .duvet/config.toml; semantics: design/duplicates/spec.md §4):
  claims.test = 1 (default)
  claims.implementation = 2
  claims.types = no free form shares a claim (default)
  targets.count = 1
  ...
```
2. **The fan-in listing names the knob in context** — when fan-in is shown and no target gate is configured, one line points at `[duplicates.targets] count / sections / types`. The reader learns the bound exists at the exact moment they're looking at what it bounds.
3. **`duvet init` writes the `[duplicates]` section into new configs**, commented with defaults shown — the feature exists in the file even when unset.

design/duplicates/decisions.md
```diff
@@ -0,0 +1,1152 @@
+# Duplicate Annotations — Decisions
+
+## Context
+
+The duplicates check today classifies each annotation type against
```
ajrudzitis, line 5: This context should be in the PR.

design/duplicates/decisions.md
```diff
@@ -0,0 +1,1152 @@
...
+
+Two boundary conditions on the model, stated so nobody over-reads
+it. First, "claim class" is well-defined only if mutual coverage
+is an equivalence relation; that holds for exact/full coverage
+and must be pinned against the code, and partial overlap (the
+`some_overlap` bucket) remains its own pathology entirely outside
+this framework. Second, an environmental fact about targets: code
+formatters relocate comments, so annotations can land on lines the
+author did not choose — same-target collisions can be tooling
+artifacts.
```
ajrudzitis, line 85: I feel like this section could really use a plain English version.  Maybe a real-life example.

design/duplicates/spec.md
```diff
@@ -0,0 +1,580 @@
...
+normalized quote): the specification section the annotation
+names, plus its quoted requirement text after whitespace
+normalization.
+
+Two annotations **share a claim** if and only if their target
+sections are identical and their normalized quotes are identical.
+
+A **claim class** is a maximal set of annotations sharing a claim,
+taken across all claim forms. A **duplicate set** is the
+restriction of a claim class to a single claim form.
```
ajrudzitis, line 59: Why is it called `form` and not `type`?

design/duplicates/decisions.md
```diff
@@ -0,0 +1,1152 @@
...
+**Context:** How does a user opt into allowing duplicates, and how
+is the legacy strict behavior preserved?
+
+### Option A: A mode flag (allow-duplicates, old/new behavior versions)
+
+- Con: a binary gate admits unbounded multiplicity the moment it
+  opens.
+- Con: mode flags fork the check's semantics permanently.
+
+### Option B: Per-type numeric caps, default 1
```
ajrudzitis, line 140: I think this is missing a major con, which is if you've raised the limit to 2 to get around one place where you need to duplicate, you now allow unnecessary duplicates universally. Seems there would be a benefit to just allowing a one-off exception.

design/duplicates/decisions.md
```diff
@@ -0,0 +1,1152 @@
...
+
+- Claim predicates: global → spec → section, keyed in the same
+  `spec.md#section-id` form annotations use — same resolver, and a
+  key naming a nonexistent section fails like a bad annotation
+  reference.
+- Target predicates: global → file path, keyed by the target's
+  own address.
+
+```toml
+[duplicates.claims."rfc9000.md#packet-format"]
```
ajrudzitis, line 942: I think there is an argument to bring forward the implementation and only allow it for claims.

design/duplicates/spec.md
```diff
@@ -0,0 +1,580 @@
...
+on its own; a free form never shares a claim with another form.
+An empty family (`types = []`) admits no multi-form set: it
+forbids all multi-form claim sharing, `test` + `implementation`
+included. Derived verdicts under the default family, for the
+record:
+
+| Coexistence (same claim) | Verdict |
+|---|---|
+| `exception` + `test` | fail — contradiction |
+| `exception` + `implementation` | fail — contradiction |
```
ajrudzitis, line 220: I almost feel like the possibility of this contradiction is a stronger justification for this type of query tool than the duplication of claims.

design/duplicates/spec.md
```diff
@@ -0,0 +1,580 @@
...
+### 4.3 Defaults {#defaults}
+
+| Setting | Default | Meaning at default |
+|---|---|---|
+| `claims.test` | 1 | test duplicate sets are singletons |
+| `claims.implementation` | 1 | implementation duplicate sets are singletons |
+| `claims.types` | all form sets containing no free form | free forms exclusive; `test`/`implementation` may share a claim |
+| `targets.count` | unlimited | listing only. Confirmed by the corpus fan-in survey: the legitimately dense population is large (s2n-quic alone bears 149 multi-annotation targets, 113 of them at count 2), so a biting default fails real dense code on day one |
+| `targets.sections` | unlimited | listing only |
+| `targets.types` | all form sets not containing both `test` and `implementation` | mixed test/implementation targets fail |
```
ajrudzitis, line 445: This seems contradict with line 433, unless I'm misunderstanding. `targets.types` reads to me like it should be the combinations that are allowed rather than not allowed based on the example above.

design/duplicates/decisions.md
```diff
@@ -0,0 +1,1152 @@
...
+
+The one composition gap, documented rather than plumbed: a project
+that raises a cap and never runs coverage holds unbilled
+duplicates. The report wording points at running both checks.
+
+---
+
+## Decision 6: Evidence units come from the artifact, never from inference {#decision-6}
+
+**Context:** Under one merged runtime report, two genuinely
```
ajrudzitis, line 446: I need help understanding the problem here.

ajrudzitis: I think the problem is that up to here you've talked about claims, types, and targets. This witness set sounds like a new dimension to the problem.

design/duplicates/decisions.md
```diff
@@ -0,0 +1,1152 @@
...
+one evidence unit there is **no execution gradient** — an
+annotation on a function head and an annotation on the exact
+enforcing line inside it bind identically under every witness that
+runs the function. No evidence-side analysis can grade *placement
+precision*; only the artifact's own structure (prover clause
+spans) ever subdivides a unit.
+
+---
+
+## Decision 7: Executed-coverage mode carries no duplicate guarantees {#decision-7}
```
ajrudzitis, line 501: I'm not here how this ties in to deduplicaton.

ajrudzitis (changes requested):
The pull request description needs to be rewritten to describe what this change actually adds to the product. Right now it feels like a summary of the discussion some AI agents had amongst themselves. 

I recommend treating the pull request description as a human-facing design document which explains what the problem is and what is being accomplished. 

(I didn't get to the code yet; still making sure I understand what is going on.)

